PR #13, opened 2018-08-23
Helm 2.10 and fix permissions

ElfoLiNk:
Updated Helm to 2.10.0

Fixed an issue when running helm init inside a container with uid:gid instead of a username:
```
docker run --rm -it -u 112:117 gcr.io/kubernetes-charts-ci/chart-testing:v1.0.3 sh 
/ $ helm init
Creating /.helm
Error: Could not create /.helm: mkdir /.helm: permission denied
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,8 @@ RUN apk --no-cache add \
     py-crcmod \
     py-pip
 
+ENV HOME /tmp
+
 # Install YQ command line reader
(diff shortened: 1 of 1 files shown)


ElfoLiNk:
@unguiculus can you have a look?

mattfarina:
@ElfoLiNk Thanks for the PR and the permission fix.

PR #14, opened 2018-08-28
[fix]: Limits release name to 29 characters.

rimusz:
Fixes an issue https://github.com/helm/charts/pull/7320#issuecomment-416213125

### DIFF
--- a/lib/chartlib.sh
+++ b/lib/chartlib.sh
@@ -376,13 +376,12 @@ chartlib::install_chart_with_all_configs() {
     release=$(yq -r .name < "$chart_dir/Chart.yaml")
 
     local random_suffix
-    random_suffix=$(tr -dc a-z0-9 < /dev/urandom | fold -w 16 | head -n 1)
-
-    local release_trimmed
-    release_trimmed=$(echo "$release"-"$random_suffix" | cut -c-29)
+    # Generate suffix 10 long as 16 long was causing StatefulSet with long names to create pods
+    # bug https://github.com/kubernetes/kubernetes/issues/64023
+    random_suffix=$(tr -dc a-z0-9 < /dev/urandom | fold -w 10 | head -n 1)
 
     local namespace="${BUILD_ID:-"$release"}-$random_suffix"
-    local release="$release_trimmed"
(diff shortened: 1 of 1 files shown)


lib/chartlib.sh
```diff
@@ -378,8 +378,11 @@ chartlib::install_chart_with_all_configs() {
     local random_suffix
     random_suffix=$(tr -dc a-z0-9 < /dev/urandom | fold -w 16 | head -n 1)
 
+    local release_trimmed
+    release_trimmed=$(echo "$release"-"$random_suffix" | cut -c-29)
```
mattfarina: Namespaces can be 63 characters long. Why the cut at 29?

rimusz: check the helm/charts#7320 issue happened

rimusz:
```
$ helm ls
artifactory-ha-7219pws7otxtcr6g
$ kubectl describe -n artifactory-ha-dmn0miwywb4tcegj statefulset
...
  Warning  FailedCreate      3m (x25 over 8m)  statefulset-controller  create Pod artifactory-ha-dmn0miwywb4tcegj-artifactory-ha-member-0 in StatefulSet artifactory-ha-dmn0miwywb4tcegj-artifactory-ha-member failed error: Pod "artifactory-ha-dmn0miwywb4tcegj-artifactory-ha-member-0" is invalid: metadata.labels: Invalid value: "artifactory-ha-dmn0miwywb4tcegj-artifactory-ha-member-545fd89668": must be no more than 63 characters
...
```

mattfarina: Wouldn't this be better if the `namespace` were a shorter name rather than things within that namespace? It would make identifying links easier.

rimusz: `namespace` is not the problem there, StatefulSet pod label `controller-revision-hash:` gets 65 characters long if there is no 29 characters limit put in for the release name.

rimusz: ok it is related to Kubernetes bug https://github.com/kubernetes/kubernetes/issues/64023

PR #20, opened 2018-09-11
Add --force and --chart option for install and lint

rimusz:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ Usage: chart_test.sh <options>
     --verbose         Display verbose output
     --no-lint         Skip chart linting
     --no-install      Skip chart installation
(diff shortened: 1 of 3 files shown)


rimusz:
ping @unguiculus @mattfarina

unguiculus:
The question is if such things should be additional flags or go into the configuration. I merged #23 first because it feels more consistent with the existing style. What do others think?

cc @helm/charts-maintainers @mattfarina

rimusz:
I will create another PR, there will reuse #23 logic to skip the version check. Regarding extra flags I found it easier to use them for local development to kick tests quicker, but I can also include the env vars support as well.

rimusz:
new #24 replaces this one, closing

Issue #29, opened 2018-09-25
Proposal: Re-implement in Go


unguiculus:
I'm tinkering with re-implementing chart-testing in Go. I've already thought about this when I wrote it in Bash and, looking at it in hindsight, I think I should have done it right from the start. I probably didn't expect we'd end up with a script containing more than 500 lines of code. Too much for a Bash script IMHO. On the other hand, Python is excellent for scripting, you can debug it, and IDE support is very good. It's easier to write and maintain than Bash.

The project is getting more popular with more PRs and issues coming in. However, adding new features to a Bash script can get messy. There are too many different styles and it's too easy to make mistakes, especially when you don't have a debugger. 

Configuration is currently not ideal. We have a mixture of CLI args and environment variables. Config and args parsing would be much easier in Python. Without any special effort, we could support CLI args, env vars, and config files.

Overall, we'd get a better understandable and maintainable code and probably lower the threshold for contributions.

Opinions?

cc @helm/charts-maintainers 

----

**Update:** The re-implementation will happen in Go!

rimusz:
yes, totally makes sense 👍

scottrigby:
Summarizing from from chat, the idea of possibly using Go. For reasons. But I don't have a strong opinion on this - IMO it's up to your preference @unguiculus 🤩

For the compile step, I reco [multistage Docker files](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds). For distribution, would love to use [goreleaser](https://github.com/goreleaser/goreleaser) (that can also solve the question of [committing the vendor dir with dep](https://github.com/scottrigby/example-dep-no-vendor)).

Minor consideration: I could pitch in more quickly if this were Go than Python 🤷‍♂️😄

rimusz:
++1 for go

unguiculus:
Initially, I chatted with @mattfarina whether Python or Go should be used. We leaned towards Python. But the more I think about it the more I'd go with Go 😉. So, Go will be the winner.

@scottrigby Im sure we'll find something for you to pitch in. 😀

prydonius:
In the charts chat, I came to the conclusion that Go might not be the best language for this, given it's going to be a lot of `exec.Command`s, it might be much more readable in something like Python. At the end of the day, as @mattfarina said, we should choose whatever language the people working on this will be more comfortable in.

unguiculus:
[...]

PR #32, opened 2018-09-28
Return correct exit code when using values files

dcarley:
The return values from the individual calls to `lint_chart_with_single_config`
are masked by the other operations in the `lint_chart_with_all_configs`
function.

When `helm lint` reports an error to STDERR but the chart is incorrectly
given a success tick and the exit code of `chart_test.sh` is 0 for
success. An example using `quay.io/helmpack/chart-testing:v1.1.0`:

    ==> Linting charts/nomad-server
    [ERROR] templates/: render error in "nomad-server/templates/nomad-server.yaml": template: nomad-server/templates/nomad-server.yaml:55:28: executing "nomad-server/templates/nomad-server.yaml" at <{{template "fullname...>: template "fullname" not defined

    Error: 1 chart(s) linted, 1 chart(s) failed
    --------------------------------------------------------------------------------
     ✔︎ charts/nomad-server
    --------------------------------------------------------------------------------
     dcarley@cci-mbp  ~/p/g/s/g/c/circle-external-services-chart   master  echo $?
    0s

Fix this by copying the pattern used in other functions of tracking
failures in a local boolean variable and using that to determine the
return code at the end of the function. This is the result after:

    ==> Linting charts/nomad-server
    [ERROR] templates/: render error in "nomad-server/templates/nomad-server.yaml": template: nomad-server/templates/nomad-server.yaml:55:28: executing "nomad-server/templates/nomad-server.yaml" at <{{template "fullname...>: template "fullname" not defined

    Error: 1 chart(s) linted, 1 chart(s) failed
    --------------------------------------------------------------------------------
     ✖︎  charts/nomad-server
    --------------------------------------------------------------------------------
     ✘ dcarley@cci-mbp  ~/p/g/s/g/c/circle-external-services-chart   master  echo $?
    1

There are probably other bugs like this. Shell scripting does not
provide safe error handling for these cases and it's difficult to write
unit tests for. This is another good reason why the proposal in #29 to
re-implement in another language is a good idea.

### DIFF
--- a/lib/chartlib.sh
+++ b/lib/chartlib.sh
@@ -380,15 +380,20 @@ chartlib::get_pods_for_deployment() {
 #   $1 The chart directory
 chartlib::lint_chart_with_all_configs() {
     local chart_dir="${1?Chart directory is required}"
+    local error=
 
     local has_test_values=
     for values_file in "$chart_dir"/ci/*-values.yaml; do
         has_test_values=true
-        chartlib::lint_chart_with_single_config "$chart_dir" "$values_file"
+        chartlib::lint_chart_with_single_config "$chart_dir" "$values_file" || error=true
     done
 
     if [[ -z "$has_test_values" ]]; then
-        chartlib::lint_chart_with_single_config "$chart_dir"
+        chartlib::lint_chart_with_single_config "$chart_dir" || error=true
+    fi
+
+    if [[ -n "$error" ]]; then
+        return 1
     fi
 }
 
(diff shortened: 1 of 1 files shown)


unguiculus:
Thanks for fixing this. I basically fixed the same thing a while ago in #18 for the install part and didn't see the same problem existed for the linting part. You fixed the problem a little differently than I did. I would like to have this done consistently in both places. Would you mind updating the PR? I don't care whether you choose your way or mine, but it should be solved the same way in both places. Thanks.

rimusz:
yes either way is fine, it just needs to be consistent

dcarley:
No problem. I've gone with the `|| error=true` pattern since that was already used in a few other places.

PR #35, opened 2018-10-19
Re-write it in Go

unguiculus:
I built a first beta release on my own fork. For this initial release I put the focus on more or less keeping the same functionality, but configuration is much more flexible now. Once we get this merged, we can look at new features such as integrating additional tools like Kubeval or better reporting capabilities.

Please review and test.

https://github.com/unguiculus/chart-testing/releases/tag/v2.0.0-beta.0

@helm/charts-maintainers

Issue #3, opened 2018-06-29
How to pass bash array env vars


hoesler:
> Note that CHART_DIRS, EXCLUDED_CHARTS, and CHART_REPOS must be configured as Bash arrays.

I cannot figure out, how to make a bash array env var accessible by the script. From what I read [here](https://unix.stackexchange.com/questions/353235/work-around-to-storing-array-values-in-an-environment-variable-then-calling-from) for example, this doesn't seem to be possible. This little example supports this assumption:

```
bash-4.4# export CHART_REPOS=("a=http://foo:8080" "b=http://bar.com:8080")
bash-4.4# echo "${CHART_REPOS[*]}"
a=http://foo:8080 b=http://bar.com:8080
bash-4.4# bash -c 'echo "${CHART_REPOS[*]}"'

bash-4.4#
```

unguiculus:
I'll try and see if I can find a solution for this. In the meantime, I'd suggest you create a config file instead of using environment variables and source it. Have a look at https://github.com/kubernetes-helm/chart-testing/blob/master/chart_test.sh for reference.

nvtkaszpir:
AFAIR the only way to export arrays in bash is to use separate file and then importing that file in subshell.

philoserf:
Should this close as #35 (go rewrite) has been merged?

hoesler:
I am fine with that.

PR #36, opened 2018-10-29
v1: Allow to pass HELM_EXTRA_OPTS in case of more complex use case scenarios

nvtkaszpir:
Please read the changes, it is explained in more detail there about the use cases.


lib/chartlib.sh
```diff
@@ -52,6 +53,11 @@ echo " CHART_YAML_SCHEMA=$CHART_YAML_SCHEMA"
 echo " VALIDATE_MAINTAINERS=$VALIDATE_MAINTAINERS"
 echo " GITHUB_INSTANCE=$GITHUB_INSTANCE"
 echo " CHECK_VERSION_INCREMENT=$CHECK_VERSION_INCREMENT"
+if [ -n "$HELM_EXTRA_OPTS" ]; then
```
unguiculus: Be consistent. Just print it as for the other vars.

unguiculus: For consistency, make it an array.

unguiculus:
FWIW, chart-testing is currently being rewritten in Go and should hopefully be merged soon. I've added this feature. See #35.

lib/chartlib.sh
nvtkaszpir: `helm test` and `helm delete` - your're right - totally forgot about that (facepalm).

About printing HELM_EXTRA_OPTS in console:
For security it was not set to be printed in the console, because someone could pass custom var override (such as password/key) - of course this could be mitigated by ensuring someone is generating full .yaml and then uses `-f` flag to override those values.

About converting HELM_EXTRA_OPTS into array:
This is explained in the expanded README.md Making it as plain string and not an array was done on purpose - not everyone is on bash 4.4 yet, also I have not checked if you can easily pass bash array as env var into the docker container. See #3

nvtkaszpir:
> FWIW, chart-testing is currently being rewritten in Go and should hopefully be merged soon. I've added this feature. See #35.

Yes I'm totally aware of this, but before it is fully merged it may take some time, while just updating current bash scripts was much quicker and easier :)

scottrigby:
Even before the documentation is updated per #45, you may start the process for https://github.com/helm/chart-testing/pull/36 by changing the base branch to be against the new `release-v1` branch. See https://help.github.com/articles/changing-the-base-branch-of-a-pull-request/ for info on how to do that with GitHub.

I will do this for now in this PR, to help ease the transition.

lib/chartlib.sh
unguiculus: Arrays are used already and Bash 4.4 is required anyways.

philoserf:
Should this close as #35 (go rewrite) has been merged?

scottrigby:
@philoserf I don't think so… per https://github.com/helm/chart-testing/pull/47/files we decided on 3 month support for each previous MAJOR version. This PR is still valid until Feb 7 2019 (Since `v2.0.0` Go rewrite was released [Nov 7 2018](https://github.com/helm/chart-testing/releases/tag/v2.0.0)). This hasn't been merged yet only because the requested changes in this discussion haven't been fully addressed: https://github.com/helm/chart-testing/pull/36#discussion_r230868957.

nvtkaszpir:
Sorry I don't have time to fix PR for at least 1 month. So I guess its better accept it or decline it.
Either way people should focus in newer version.

PR #50, opened 2018-11-10
Bind debug flag to Viper

unguiculus:
The flag wasn't bound to Viper. Setting it would have no effect.


rimusz:
LGTM

PR #56, opened 2018-11-16
Explicitly disable version checks when all or specific charts are tested

unguiculus:
It is already documented that the check is disabled in this case.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -114,6 +114,11 @@ func LoadConfiguration(cfgFile string, cmd *cobra.Command, bindFlagsFunc ...func
 		cfg.LintConf = cfgFile
 	}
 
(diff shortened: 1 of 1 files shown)


rimusz:
awesome, thanks

rimusz:
now bash and go based chart-testing work in pair

onedr0p:
@unguiculus (fast forward 2.3 years) 

I am curious about this logic, since you have the `check-version-increment` flag why is this even needed? It's not really UX friendly, we're trying to build a workflow where we matrix out chart changes and it's currently working as expected but not what we expect.

https://github.com/k8s-at-home/charts/blob/8ec4a9ab93fab9f2952dc25f9898e4607ffb5be4/.github/workflows/charts-lint-test.yaml#L157-L159

PR #59, opened 2018-11-17
Support installation to existing namespace

### LINKED ISSUE #34: Support chart installation to an existing namespace
I am working with an organization that runs a dedicated tiller instance for each set of namespaces owned by a given business unit. None of these tiller instances have cluster-admin permissions, nor can they create/delete arbitrary namespaces. Teams may also maintain their own chart repositories.

Specifying an existing namespace for chart installation would allow teams to consume this library in a more distributed and self-service manner, without the need for a dedicated cluster or access to a more highly privileged tiller instance.

jlegrone:
This PR adds two flags to the `install` command: `namespace` and `release-label`. If `namespace` is specified, releases will target that namespace and `release-label` will be used to select deployments and pods for readiness and reading log output.

Fixes #34

I'm marking this as a WIP to gather feedback on the design (and whether we want to enable this functionality -- I wasn't totally clear on that based on the discussion so far in #34).  I'll be happy to add more tests for this change if desired. 😄


pkg/chart/chart.go
```diff
@@ -71,6 +72,7 @@ type Helm interface {
 	LintWithValues(chart string, valuesFile string) error
 	Install(chart string, namespace string, release string) error
```
jlegrone: This method is now unused; we may want to remove this or refactor to replace `InstallWithValues`.

pkg/config/config.go
```diff
@@ -52,6 +53,8 @@ type Configuration struct {
 	ExcludedCharts        []string `mapstructure:"excluded-charts"`
 	HelmExtraArgs         string   `mapstructure:"helm-extra-args"`
 	Debug                 bool     `mapstructure:"debug"`
+	Namespace             string   `mapstructure:"namespace"`
+	ReleaseLabel          string   `mapstructure:"release-label"`
```
jlegrone: TODO: Add check for zero value for `ReleaseLabel` if `Namespace` is set.

jlegrone: Also, this option can probably be removed in v3 when we can just look for resources that are owned by the helm release CRD:

https://github.com/helm/community/blob/master/helm-v3/003-state.md#the-release-object

unguiculus:
I will need some time to review this. Please bear with me.

pkg/chart/chart.go
```diff
@@ -166,7 +168,7 @@ func NewTesting(config config.Configuration) Testing {
 	extraArgs := strings.Fields(config.HelmExtraArgs)
 	testing := Testing{
 		config:           config,
-		helm:             tool.NewHelm(procExec, kubectl, extraArgs),
+		helm:             tool.NewHelm(procExec, extraArgs),
```
unguiculus: The `kubectl` variable can be inlined now.

pkg/chart/chart.go
```diff
@@ -326,28 +328,35 @@ func (t *Testing) InstallChart(chart string, valuesFiles []string) TestResult {
...
-		defer t.kubectl.DeleteNamespace(namespace)
 		defer t.helm.DeleteRelease(release)
-		defer t.PrintPodDetailsAndLogs(namespace)
 
-		if err := t.helm.Install(chart, namespace, release); err != nil {
+		if t.config.Namespace != "" {
+			namespace = t.config.Namespace
+			releaseSelector = fmt.Sprintf("%s=%s", t.config.ReleaseLabel, release)
+		} else {
+			defer t.kubectl.DeleteNamespace(namespace)
```
unguiculus: The problem now is that the cleanup order is no longer correct because you call this after `defer t.helm.DeleteRelease(release)`. Please keep the old order.

pkg/chart/chart.go
unguiculus: I'd say remove and refactor

unguiculus: We should use the same logic for `Lint`.

pkg/chart/chart.go
```diff
@@ -326,28 +327,38 @@ func (t *Testing) InstallChart(chart string, valuesFiles []string) TestResult {
...
+	}
+
+	for _, valuesFile := range valuesFiles {
+		var namespace, release, releaseSelector string
 
+		if t.config.Namespace != "" {
+			namespace = t.config.Namespace
+			release, _ = util.CreateInstallParams(chart, t.config.BuildId)
+			releaseSelector = fmt.Sprintf("%s=%s", t.config.ReleaseLabel, release)
+			defer t.helm.DeleteRelease(release)
```
unguiculus: `defer t.helm.DeleteRelease(release)` could be moved to after the `if/else` block.

pkg/chart/chart.go
jlegrone: I've removed `Lint` and `Install`. That seems like the smallest possible change, but we could also replace these function implementations (e.g. rename `InstallWithValues` --> `Install`) instead.

As an aside, we might consider moving `Helm`, `Kubectl`, etc interfaces and implementations to internal packages in order to avoid breaking changes when doing these sorts of refactors in the future.

PR #62, opened 2018-11-28
Reduce Docker Image Size

philoserf:
Size

- use the _builder_ pattern
- use the _RUN ... \ && ..._ pattern
- use _COPY item1 item2 ... destination_ pattern
- use a python/alpine image rather than installing python, etc
- remove apk packages that are not required
- remove cache/control file from final image layer

Other Changes

- use cURL flags -sLo to improve output and control local file names

Signed-off-by: Mark Ayers <mark@philoserf.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,41 +1,37 @@
-FROM alpine:3.8
+## BUILDER
+#
+FROM alpine:3.8 as build
 
-RUN apk --no-cache add \
-    curl \
-    git \
-    libc6-compat \
-    openssh-client \
-    python \
-    py-crcmod \
-    py-pip && \
-    pip install --upgrade pip==18.1
-
-# Install a YAML Linter
-ARG YAML_LINT_VERSION=1.12.1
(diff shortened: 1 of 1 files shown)


philoserf:
FWIW: v2.0.1 was 207MB This is 175MB. It should also be a bit faster at build time.

philoserf:
It occurred to mean as I slept that perhaps `git` was used in the `ct` workflow. Taking a closer look at how the tool is used.

unguiculus:
Thanks for your contribution. What makes you think some packages are no required? If you add them back you probably won't save anything by restructuring the Dockerfile. I don't think it make sense to use a multi-stage docker build here. The original Dockerfile is IMO clearer and better to understand.

philoserf:
It is easier to read. Granted.

Size, layer count, and speed can matter. I propose here some of Docker’s own best practices for Dockerfiles.

philoserf:
BTW. I took a look into the image with `dive`. This lead me to some of the size savings. 

Ref: https://github.com/wagoodman/dive/blob/master/README.md

PR #63, opened 2018-11-28
Remove external heredoc library dependency

philoserf:
- go already provides the capability natively with multiline strings
- incidental reordering of imports by `goimports` on edited files

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,14 +2,6 @@
 
 
 [[projects]]
-  branch = "master"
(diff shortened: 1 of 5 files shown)


unguiculus:
Multiline strings are already used. Heredoc adds support proper indentation.

philoserf:
While true it was using multi line strings within heredoc, I guess this pull request expresses my opinion about adding an external dependency. I believe the heredoc way isn’t better and costs more. 

I can withdraw the PR as my commit statement is incomplete or we can consider it and I will update the commit message.

PR #67, opened 2018-12-15
Add support for providing credentials when adding chart-repositories

Flydiverny:

Add support for providing chart repo credentials when adding private chart repos

First time writing go, So be nice <3


Flydiverny:
Seems like a nicer solution! I pushed a variant with a `helm-repo-extra-args` config.
I did pretty much the same as the first solution to allow for passing different args for different repositories.

so you can do something like this

```sh
ct lint --chart-repos "private-charts=https://charts.private.com,ambassador=https://www.getambassador.io" --helm-repo-extra-args "gleerups-charts=--username user --password secret,ambassador=--no-update"
```

Or yaml config:
```yaml
chart-repos:
  - incubator=https://incubator
  - private=https://private
helm-repo-extra-args:
  - private=--username test --password secret
```

pkg/chart/chart.go
```diff
@@ -197,12 +197,23 @@ func (t *Testing) processCharts(action func(chart string, valuesFiles []string)
 		return nil, errors.Wrap(err, "Error initializing Helm")
 	}
 
+	repoArgs := map[string][]string{}
```
unguiculus: I'd say you do too much here. Please use the same logic as for `helm-extra-args` and pass the args into the `Helm` struct.

See https://github.com/helm/chart-testing/blob/master/pkg/chart/chart.go#L163

Flydiverny: Not sure I follow, do you agree with the way of providing different args for different repos? And the logic for parsing the args move to NewHelm with AddRepo internally select the correct args when given a repo to add?

unguiculus: I guess I should have look more carefully, sorry. I'll have another look at this.

app/cmd/root.go
```diff
@@ -77,6 +77,9 @@ func addCommonLintAndInstallFlags(flags *pflag.FlagSet) {
 	flags.StringSlice("chart-repos", []string{}, heredoc.Doc(`
 		Additional chart repos to add so dependencies can be resolved. May be
 		specified multiple times or separate values with commas`))
+	flags.StringSlice("helm-repo-extra-args", []string{}, heredoc.Doc(`
+		Additional arguments for Helm. (e.g. --username test --password secret)
```
unguiculus: Please state that it's specifically for the `helm repo` command.

pkg/chart/chart.go
```diff
@@ -197,12 +197,23 @@ func (t *Testing) processCharts(action func(chart string, valuesFiles []string)
...
 	for _, repo := range t.config.ChartRepos {
 		repoSlice := strings.SplitN(repo, "=", 2)
 		name := repoSlice[0]
 		url := repoSlice[1]
-		if err := t.helm.AddRepo(name, url); err != nil {
-			return nil, errors.Wrapf(err, "Error adding repo: %s=%s", name, url)
+
+		repoExtraArgs := repoArgs[name]
+		if err := t.helm.AddRepo(name, url, repoExtraArgs); err != nil {
+			return nil, errors.Wrapf(err, "Error adding repo: %s=%s with args %s", name, url, repoExtraArgs)
```
unguiculus: Don't include the args in the error message. We don't want to log sensitive data.

app/cmd/root.go
```diff
@@ -77,6 +77,9 @@ func addCommonLintAndInstallFlags(flags *pflag.FlagSet) {
 	flags.StringSlice("chart-repos", []string{}, heredoc.Doc(`
 		Additional chart repos to add so dependencies can be resolved. May be
 		specified multiple times or separate values with commas`))
+	flags.StringSlice("helm-repo-extra-args", []string{}, heredoc.Doc(`
+		Additional arguments for helm repo add. (e.g. --username test --password secret)
+		May be specified multiple times or separate values with commas`))
```
unguiculus: Looking at this again, we should make clear how to configure it, i. e. with `--helm-repo-extra-args <repo>=<args>`. An example in the readme would be great as well.

unguiculus: How about this:

```go
flags.StringSlice("helm-repo-extra-args", []string{}, heredoc.Doc(`
	Additional arguments for the 'helm repo add' command to be 
	specified on a per-repo basis with an equals sign as delimiter 
	(e. g. 'myrepo=--username test --password secret'). May be specified
	multiple times or separate values with commas`))
```

Flydiverny:
@unguiculus pr updated and rebased :)

README.md
```diff
@@ -95,6 +95,23 @@ build-id: pr-42
...
+```yaml
+chart-repos:
+  - incubator=https://incubator.io
+  - basic-auth-repo=https://private.com
+  - ssl-repo=https://self-signed.ca
+helm-repo-extra-args:
+  - ssl-repo=--ca-file ./my-ca.crt
+```
+
+    ct install --config config.yaml ---helm-repo-extra-args "basic-auth=--username user --password secret"
```
unguiculus: One hyphen too many.

```suggestion
    ct install --config config.yaml --helm-repo-extra-args "basic-auth=--username user --password secret"
```

Flydiverny: ops, good catch. pr updated.

Issue #53, opened 2018-11-14
Remove implicit dependence on extra tools/repos


munnerz:
Hey all 👋 

First of all, thanks very much for this project. We use the old bash version of this in cert-manager and it works well 😄 

The recent Go rewrite is great, and I'm keen to get this setup in our own environment again.

I wonder if there is any desire to remove the dependence on yamale/yamllint?

We're hoping to be able to run this without the need for Docker, so for the time being will need to configure our own build/test environment to grab the appropriate version of these tools.

Can these projects either be included in the Go binary release somehow, or otherwise can their functionality be absorbed into Go?

unguiculus:
Hey @munnerz, great to hear you use it. 😄

I'm not aware of any alternatives in Go that could replace these tools. However, I guess we could add flags for disabling them.

cpanato:
The flags will be set up in the CT_XXXX env flags? 
I would like to try out this one. @unguiculus can we define the scope?

thanks

unguiculus:
@cpanato You just flags in `lint.go`. The user is then free to choose whether they want to use the cli flag, an env var, or a config file to configure it.

I would add two flags:
* `no-chart-schema-validation`
* `no-yaml-lint`

What do you think?
@munnerz

PR #68, opened 2018-12-18
Add new flags to disable lint/schema validation

cpanato:
Add new flags to by pass lint/schema validation: `no-chart-schema-validation` and `no-yaml-lint`

**Which issue this PR fixes** 
  - fixes https://github.com/helm/chart-testing/issues/53


pkg/chart/chart.go
```diff
@@ -280,17 +280,22 @@ func (t *Testing) LintChart(chart string, valuesFiles []string) TestResult {
...
+		if err := t.linter.Yamale(chartYaml, t.config.ChartYamlSchema); err != nil {
+			result.Error = err
+			return result
+		}
 	}
-	if err := t.linter.YamlLint(valuesYaml, t.config.LintConf); err != nil {
-		result.Error = err
-		return result
+
+	if !t.config.NoYamlLint {
```
unguiculus: Reflecting about this again, I'm think it would be better to do it the other way around an call the flags `validate-yaml` and `validate-chart-schema`. This would also be more consistent with the existing flags.

cpanato: sure, was hard to me to think in the negative 😄 i will update
thanks for the review @unguiculus

cpanato: done

unguiculus:
I added some minor updates if you don't mind.

cpanato:
@unguiculus no problem from my side. thanks btw!

PR #73, opened 2019-01-02
Fix index out of range when files exist in the root of a chart dir

### LINKED ISSUE #72: index out of range if there are files in the root of the repository
BUG REPORT

If the root of a repository is used as the chart dir (`.`) and there are non-empty directories or files, an index out of range will be raised.

Attaching a PR that reproduces and fixes the issue.

kunickiaj:
If the root of a repository is used as the chart dir (`.`) and there are non-empty directories or files, an index out of range will be raised.

fixes #72

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -408,7 +408,7 @@ func (t *Testing) ComputeChangedChartDirectories() ([]string, error) {
 	var changedChartDirs []string
 	for _, file := range allChangedChartFiles {
 		pathElements := strings.SplitN(filepath.ToSlash(file), "/", 3)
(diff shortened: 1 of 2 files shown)


unguiculus:
Thanks for the PR. While this fixes the error, it fails to recognize any chart changes at all. The current logic is not built for the case that a chart dir is at the repo root, which is clearly an oversight. I'll try and look into this as well.

unguiculus:
OK, I looked into this a little bit deeper. I think I misunderstood. Your fix works as intended. I had initially thought you had the chart directory itself, i. e. the `Chart.yaml` itself in the repo root, which is not supported.

unguiculus:
LGTM

unguiculus:
@ZachEddy I hope to create a release next week.

PR #76, opened 2019-01-07
Update Goreleaser

unguiculus:


scottrigby:
For reference later, this was due to https://github.com/goreleaser/goreleaser/pull/896

PR #82, opened 2019-01-14
--chart-repos help message improvement

depohmel:
What this PR does / why we need it:
pkg/chart/chart.go:201 is expecting repo to be name=url
this PR is just making help message clear


unguiculus:
LGTM

PR #94, opened 2019-01-24
Update kind example

unguiculus:
* Use kind release from GitHub instead of 'go get sigs.k8s.io/kind'
* Run ct container with '--network host' to avoid patching kubeconfig
* Create multi-node cluster with one master and three worker nodes
  so pods with anti-affinity can be tested
* Improve wait logic to cater for multiple nodes
* Add log statements
* Remote unnecessary 'k8s' remote to avoid confusion
* Extract function for 'docker exec' calls

### DIFF
--- a/examples/kind/.circleci/config.yml
+++ b/examples/kind/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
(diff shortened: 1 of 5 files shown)


unguiculus:
I did test it pretty extensively. 🤓

PR #103, opened 2019-02-12
Test chart upgrades

### LINKED ISSUE #19: Check for mutating matchLabels in apps/v1
`matchLabels` cannot change in `apps/v1` (e.g., StatefulSet). That means things like the `chart` label cannot change when the version increments. This is no longer allowed.

A way to catch an error on the chart or other common labels would be useful.

### LINKED ISSUE #25: Automate tests to upgrade individual charts
Community charts were hit with a set of unexpected upgrade bugs. We can speculate about why that didn't come to our attention before, but this also could have been caught (and will catch future issues) by adding automated upgrade tests. 

See:
- https://github.com/helm/charts/issues/7578
- https://github.com/helm/charts/issues/7680
- https://github.com/helm/charts/issues/7726

### LINKED ISSUE #99: Check for mutating metadata.labels in v1beta1
See https://github.com/helm/charts/issues/7726

Related: #19

### LINKED ISSUE #100: Check for mutating labels in StatefulSet spec.VolumeClaimTemplate.metadata.labels
See https://github.com/helm/charts/issues/7803

Related #19, #99

jlegrone:
This PR enables testing in-place upgrades of charts that have non-breaking changes according to the [semver spec](https://semver.org).

The desired behavior of this PR and further discussion can be found here: https://github.com/helm/chart-testing/issues/25#issuecomment-447426036

**Which issue this PR fixes**

closes #25
closes #19 
closes #99 
closes #100 

Some TODOs:
- [x] Handle uncommitted changes in git repositories
- [x] Handle not being in a git repository
- [x] Add a config flag to opt out of upgrade testing
- [x] Consider the need for more tests (PR #120)
- [x] ~Change the `Git` interface to a "SCM" interface to avoid further tight coupling with git?~ (out of scope for this PR)

Known caveats:

1. Only works with git
1. If a chart has direct dependencies outside of its parent directory, `helm dependency build` may fail or not produce the same result as the previous revision

~Run integration tests with `go test -cover -tags=integration ./...`~ (see #120)


jlegrone:
@unguiculus I'd appreciate a first pass when you have a chance!

unguiculus:
Thanks. Please give me some time to review and test this. From quickly skimming over it: Have you thought about installing the previous version from the repo instead of checking out the merge base? Doing so, you assume the merge base reflects the previous release. This doesn't necessarily have to be the case.

jlegrone:
I did consider pulling the previous published revision, but I think there are a couple problems with that:

1. We don't necessarily know _where_ the charts have been published -- I suppose we could handle this with a new configuration flag.
2. Not all teams publish their charts to a registry. It is a common use case to version a chart alongside your application source code and run helm commands using the local chart path.

pkg/chart/chart.go
```diff
@@ -349,47 +418,124 @@ func (t *Testing) InstallChart(chart string, valuesFiles []string) TestResult {
...
+	if result.Error != nil {
+		// Upgrading chart from previous release failed; return result.
+		return result
+	}
+
+	return result
+}
+
+func (t *Testing) testRelease(release, namespace, releaseSelector string) error {
+	if err := t.kubectl.WaitForDeployments(namespace, releaseSelector); err != nil {
```
jlegrone: Turns out this will likely fail for deployments or statefulsets in burst mode (I was doing most of my manual testing against the redis chart which uses a statefulset). Working on a fix...

jlegrone:
Sorry about the size... I would be happy to factor the integration tests out into a separate PR.

kind-config.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Config
+apiVersion: kind.sigs.k8s.io/v1alpha2
+nodes:
+  - role: control-plane
+  - role: worker
+    replicas: 3
```
unguiculus: Use 1 replica. See https://github.com/helm/chart-testing/pull/117

pkg/chart/chart.go
```diff
@@ -42,9 +43,13 @@ import (
 type Git interface {
 	FileExistsOnBranch(file string, remote string, branch string) bool
 	Show(file string, remote string, branch string) (string, error)
+	CheckoutDir(directory string, ref string) error
+	CleanDir(directory string) error
+	IsDirClean(directory string) (bool, error)
 	MergeBase(commit1 string, commit2 string) (string, error)
 	ListChangedFilesInDirs(commit string, dirs ...string) ([]string, error)
 	GetUrlForRemote(remote string) (string, error)
+	ValidateRepository() error
```
unguiculus: Document methods

pkg/chart/chart.go
```diff
@@ -68,7 +73,8 @@ type Helm interface {
 	BuildDependencies(chart string) error
 	LintWithValues(chart string, valuesFile string) error
 	InstallWithValues(chart string, valuesFile string, namespace string, release string) error
-	Test(release string) error
+	Upgrade(chart string, release string) error
+	Test(release string, cleanup bool) error
```
unguiculus: Document methods

pkg/chart/chart.go
```diff
@@ -171,7 +178,31 @@ func NewTesting(config config.Configuration) Testing {
 		directoryLister:  util.DirectoryLister{},
 		chartUtils:       util.ChartUtils{},
 	}
-	return testing
+	if testing.config.Upgrade {
```
unguiculus: I don't like this. `NewTesting` is a constructor-like function and, as such, should not contain logic. Neither should it return an error.

ct/cmd/install.go
```diff
@@ -61,12 +61,15 @@ func addInstallFlags(flags *flag.FlagSet) {
 	flags.String("helm-extra-args", "", heredoc.Doc(`
 		Additional arguments for Helm. Must be passed as a single quoted string
 		(e.g. "--timeout 500 --tiller-namespace tiller"`))
+	flags.Bool("upgrade", true, heredoc.Doc(`
```
unguiculus: The default should be `false` in order to not break the existing behavior.

pkg/chart/chart.go
```diff
@@ -222,13 +253,42 @@ func (t *Testing) processCharts(action func(chart string, valuesFiles []string)
...
+			}
+			// Set contents of charts directory to the target branch
+			t.git.CheckoutDir(dir, t.mergeBase)
+			// Copy charts directory contents to a temp directory
+			prevRevisionDir := tempChartParentDir(dir)
+			copy.Copy(dir, prevRevisionDir)
+			// Reset charts directory to last commit
+			t.git.CheckoutDir(dir, "HEAD")
+			// Schedule cleanup of untracked temp directory
+			defer t.git.CleanDir(prevRevisionDir)
```
unguiculus: `defer` should not be used inside a loop.

jlegrone: To me the defer feels more maintainable since we schedule the cleanup action right after creating the temp directory, and we would need to do some additional bookkeeping for handling early returns otherwise.

Memory overhead here should be very low since we're only keeping track of the `prevRevisionDir` string, and disk usage scales linearly (if disk is a concern we'll need to do a little additional refactoring).

ct/cmd/install.go
jlegrone: I'm good with this, though I hope we can flip this to true in a release or two. It would be nice for end users to get all the benefits of `ct` out of the box without having to figure out which additional flags they need to enable.

unguiculus:
@jlegrone Please don't rebase and amend commits but add additional ones on top. This makes it easier to review and GitHub can show you the changes. We squash upon merge anyways.

jlegrone:
@unguiculus sure thing. For this PR I was thinking it would make sense not to squash since the integration test changes are so massive, but up to you.

I'm working on adding a bit more documentation to the install command now. Regarding this comment:

> You can configure repos with `--chart-repos`. You can probably just install the latest version from the repo. This should IMO be the default with an option to use merge base.

This would work a lot of the time, but especially as we move to a more decentralized model it won't be unheard of to have multiple custom repositories, each with different versions of the same chart name. There has also not been a requirement to add custom repositories so far, unless you are pulling dependencies in from those repositories.

For example, I could be pulling in the bitnami charts repository to take advantage of their `memcached` chart, while using `ct` to test my own `nginx` chart. Because the bitnami repository also publishes an `nginx` chart, how does `ct` tell which remote repository is correct for the chart under test?

We can definitely handle this case, and I'm interested in your thoughts on implementation. I think it would be best done as a future enhancement since this feature will likely need to introduce some new configuration options.

Aside from the above, IMO testing against merge base makes sense as the default since

1. it is in line with how we already validate version bumps in `Chart.yaml`
2. it is lower barrier to entry (doesn't require a standalone repository for publishing)
3. it is a reasonable default for teams that are practicing continuous delivery, and is the only option for teams running helm commands against a chart directory from version control

unguiculus:
Makes sense. I haven't thought this through yet. Let's stick with merge-base for now which, I agree, is a reasonable default.

Re. squashing: We don't necessarily have to squash everything. In this case, I'd suggest to make the history nice once we're ready to merge. Alternatively, you could factor out integration tests into a separate PR as you suggested earlier.

jlegrone:
@unguiculus sounds good, I'll rework the commit log once you give an initial approval and we can just do a merge commit.

unguiculus:
I talked with @scottrigby about it the other day. We agreed that it would be best to split out the integration tests into a separate PR.

jlegrone:
@unguiculus I've opened a separate PR as requested (#120). I'll keep that changeset rebased on top of this PR to verify that integration tests continue to pass.

Gopkg.lock
```diff
@@ -92,6 +92,14 @@
   revision = "3536a929edddb9a5b34bd6861dc4a9647cb459fe"
   version = "v1.1.2"
 
+[[projects]]
+  branch = "master"
+  digest = "1:a4c2e29d2fb620f39ab706071791242a25df5c62e4793f9b401b07e490d535e9"
+  name = "github.com/otiai10/copy"
```
scottrigby: Curious how the lock file changed if the toml file didn't?

ct/cmd/install.go
```diff
@@ -31,13 +31,17 @@ func newInstallCmd() *cobra.Command {
...
 
 			* changed charts (default)
 			* specific charts (--charts)
 			* all charts (--all)
 
-			in given chart directories.
+			in given chart directories. If upgrade (--upgrade) is true, then this
+			command will validate that 'helm test' passes for the following upgrade paths:
+
+			* previous chart revision => current chart version (if non-breaking SemVer change)
```
scottrigby: Should this be "previous chart revisions" plural?

pkg/chart/chart.go
scottrigby: If I understand this correctly, when using defer inside a loop we should wrap it in either an anonymous or a named function. See my note about previous version(s) above… this may cause the loop to pile up a lot more.

ct/cmd/install.go
scottrigby: Never mind, I see it is not, but see my note above – I believe it should be.

Gopkg.lock
jlegrone: I just added an import for this lib in source and I think ran `dep ensure`. Looks like maybe I should run `dep ensure -add github.com/otiai10/copy` instead?

scottrigby: Well… check out https://github.com/golang/dep/issues/1670. When all else fails, I usually do a fresh `dep init` just to see what's changed 🤷‍♂️

ct/cmd/install.go
jlegrone: Re testing _all_ previous semver-compatible revisions:

Yes this would be ideal, though I do worry about the impact this could have on test duration. Perhaps testing the very first compatible version and the very last would be a good enough signal? Can you think of any cases where this would break down?

scottrigby: An example is if someone broke then fixed one of the label immutability issues (or some other breaking change) without bumping the MAJOR version. In this case, one or more MINOR versions may fail the next upgrade even if the first and last one pass.

If we test upgrades between individual releases, issues like this should be caught moving forward… but it won't help to fix any problems that exist before we start testing upgrades between each release.

Practically speaking this may be too much for our testing large chart repos (like https://github.com/helm/charts). Good idea to meet, let's discuss Tomorrow and decide.

pkg/chart/chart.go
```diff
@@ -509,6 +684,24 @@ func (t *Testing) CheckVersionIncrement(chart string) error {
...
 }
 
+func (t *Testing) checkBreakingChangeAllowed(chart string) (allowed bool, reasons []error, err error) {
+	oldVersion, err := t.GetOldChartVersion(chart)
+	if err != nil {
+		return false, nil, err
+	}
+	if oldVersion == "" {
+		// new chart, skip upgrade check
+		return true, []error{fmt.Errorf("chart has no previous revision")}, nil
```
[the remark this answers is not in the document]
jlegrone: Agreed this is weird... this is the https://github.com/Masterminds/semver api leaking.

pkg/chart/chart.go
```diff
@@ -222,13 +268,42 @@ func (t *Testing) processCharts(action func(chart string, valuesFiles []string)
...
+			changedParentDirs[path.Dir(dir)] = true
+		}
+		for dir := range changedParentDirs {
+			// Check for uncommitted changes that would be lost when checking out older revision
+			clean, _ := t.git.IsDirClean(dir)
+			if !clean {
+				return results, fmt.Errorf("Directory %s has uncommitted changes", dir)
+			}
+			// Set contents of charts directory to the target branch
+			t.git.CheckoutDir(dir, t.mergeBase)
```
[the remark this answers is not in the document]
jlegrone: Oooh I hadn't seen that before! That looks like just what we need, and also should remove potential for some weird edge cases where charts might have dependencies outside of their parent directories.

pkg/chart/chart.go
```diff
@@ -183,6 +216,19 @@ func (t *Testing) processCharts(action func(chart string, valuesFiles []string)
 		return results, nil
 	}
 
+	if t.config.Upgrade {
```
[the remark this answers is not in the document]
jlegrone: We do need to do some stuff here though, like building dependencies for previous revisions. I think `Upgrade` will always be false if we're only linting.

pkg/chart/chart.go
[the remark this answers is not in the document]
jlegrone: This also seems better in terms of disk usage relative to recursively copying the whole chart parent directory!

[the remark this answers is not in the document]
jlegrone: This also seems better in terms of disk usage relative to recursively copying files from each chart parent directory. 😄

pkg/util/util.go
```diff
@@ -177,6 +178,30 @@ func CompareVersions(left string, right string) (int, error) {
 	return leftVersion.Compare(rightVersion), nil
 }
 
+func BreakingChangeAllowed(left string, right string) (bool, []error, error) {
```
unguiculus: I'd suggest to only return `bool` and `error`. Let's add https://github.com/hashicorp/go-multierror and wrap multiple errors.

pkg/chart/chart.go
```diff
@@ -222,6 +245,23 @@ func (t *Testing) processCharts(action func(chart string, valuesFiles []string)
 		TestResults:    results,
 	}
 
+	// Checkout previous chart revisions and build their dependencies
+	if t.config.Upgrade {
```
unguiculus: As mentioned earlier, this is also called when linting, which doesn't make sense.

pkg/chart/chart.go
```diff
@@ -336,10 +411,55 @@ func (t *Testing) LintChart(chart string, valuesFiles []string) TestResult {
...
+	var result TestResult
+
+	if t.config.Upgrade {
+		// Test upgrade from previous version
+		result = t.UpgradeChart(chart)
+		if result.Error != nil {
+			return result
+		}
+		// Test upgrade of current version (related: https://github.com/helm/chart-testing/issues/19)
+		if err := t.testUpgrade(chart, chart, true); err != nil {
```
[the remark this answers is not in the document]
unguiculus: Have you thought about this? I'd prefer to remove `test` from the method names. How about something like `doUpgrade` and `doInstall` instead?

pkg/chart/chart.go
jlegrone: Would you prefer that we implement this in `InstallCharts` and `LintAndInstallCharts` instead, or somewhere else? The reason I haven't done this was that I wanted to avoid calling `FindChartsToBeProcessed` multiple times or adding more args to `processCharts` if possible. It is also a bit more complex to do this prep in a utility function since we'd have to pass a cleanup callback function to remove the working tree rather than just using defer.

I feel like it makes sense to have a run-once "prepare" function. Here is the current behavior as I understand it:
- If `ct lint` is run, `--upgrade` is not an option and this code branch will never be executed (ct exits with `unknown flag: --upgrade`).
- If `ct lint-and-install` is run, this code branch is only executed once.

There may also come a time when we want previous chart revisions available for more than the upgrade testing use case. For example, the git tool's `FileExistsOnBranch` and `Show` methods could both be implemented using the previous revision's worktree rather than going over the network. If/when this were the case, keeping this functionality directly in `processCharts` would make that a very minimal refactor.

If you're still not convinced then no worries, please just let me know where this should be implemented. 😄

pkg/chart/chart.go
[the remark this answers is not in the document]
unguiculus: Yeah, I figured going through the code again. However, `Upgrade` can be `true` if set as env var or via config file when linting. We should explicitly set it to `false` when the config is loaded in this case below this line: https://github.com/helm/chart-testing/blob/master/pkg/config/config.go#L113

pkg/chart/chart.go
unguiculus: I commented again about this further above (https://github.com/helm/chart-testing/pull/103#discussion_r264227348), probably at the wrong place:

> Yeah, I figured going through the code again. However, `Upgrade` can be `true` if set as env var or via config file when linting. We should explicitly set it to `false` when the config is loaded in this case below this line: https://github.com/helm/chart-testing/blob/master/pkg/config/config.go#L113

pkg/chart/chart.go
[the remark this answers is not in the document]
jlegrone: I just did some testing and couldn't find a way to set `Upgrade` to true when running `ct lint`, even when making true the default value for the flag, using an environment variable, or setting it in a config file. It seems like viper does a pretty good job with this already, but also went ahead and added an explicit check just for peace of mind.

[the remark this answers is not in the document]
unguiculus: Interesting. I did test exactly that case with a config file and `Upgrade` was set to true. Will check again.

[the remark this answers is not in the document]
jlegrone: Oh no you're right! I thought I had tested with a config file yesterday -- confirmed this is the one method that works for tricking viper. Not sure why that does it but not the env variable...

unguiculus:
LGTM but @scottrigby should also approve.

ct/cmd/install.go
scottrigby: Update: we discussed this and agreed the use case I mentioned should only need to be run once per repo, if thereafter the most recent release is tested for upgrades. Therefore, if we care to do this for a historic list (to catch up on an existing repo) it could either be scripted manually, or else added as a new single-use function in a follow-up PR.

PR #111, opened 2019-02-18
force the namespace deletion by updating the finalizers in the namespace obj

cpanato:
After the namespace deletion sometimes the namespace get stuck in the `Terminating` phase forever and we need to use the script from @unguiculus (https://gist.github.com/unguiculus/d915e911ba04f1f97ecea4f617a2eef5) to remove the namespaces.

This PR introduces the same login in the bash script to the chart-testing, when we check if the namespace still exists we apply the logic to force remove the namespace.

note: for example, last week in the Helm Chart CI Cluster we had around 900 namespaces in Terminating phase.

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -96,3 +96,16 @@ func (p ProcessExecutor) RunProcess(executable string, execArgs ...interface{}) 
 
 	return nil
 }
+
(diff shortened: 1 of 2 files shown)


cpanato:
not working after some tests, will check and reopen another PR

PR #112, opened 2019-02-18
force the namespace deletion by updating the finalizers in the namespace obj

cpanato:
After the namespace deletion sometimes the namespace get stuck in the `Terminating` phase forever and we need to use the script from @unguiculus (https://gist.github.com/unguiculus/d915e911ba04f1f97ecea4f617a2eef5) to remove the namespaces.

This PR introduces the same login in the bash script to the chart-testing, when we check if the namespace still exists we apply the logic to force remove the namespace.

note: for example, last week in the Helm Chart CI Cluster we had around 900 namespaces in Terminating phase.

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -96,3 +96,16 @@ func (p ProcessExecutor) RunProcess(executable string, execArgs ...interface{}) 
 
 	return nil
 }
+
+func (p ProcessExecutor) RunLongRunningProcess(executable string, execArgs ...interface{}) (*exec.Cmd, error) {
+	args, err := util.Flatten(execArgs)
+	if p.debug {
+		fmt.Println(">>>", executable, strings.Join(args, " "))
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "Invalid arguments supplied")
+	}
+	cmd := exec.Command(executable, args...)
+
+	return cmd, nil
+}
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -121,10 +121,6 @@ func (p ProcessExecutor) RunWithProxy(withProxy fn) error {
 		return errors.Wrap(err, "Error starting the 'kubectl proxy' process")
 	}
 
-	go func() {
-		cmdProxy.Wait()
-	}()
-
 	err = withProxy(randomPort)
 
 	cmdProxy.Process.Signal(os.Kill)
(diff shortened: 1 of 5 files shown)


pkg/exec/exec.go
```diff
@@ -96,3 +96,16 @@ func (p ProcessExecutor) RunProcess(executable string, execArgs ...interface{})
 
 	return nil
 }
+
+func (p ProcessExecutor) RunLongRunningProcess(executable string, execArgs ...interface{}) (*exec.Cmd, error) {
```
unguiculus: I don't like the name of the method. `Run` is not correct because it doesn't start the process. Also, it doesn't have to be long-running.

Why not just call it `CreateProcess` and even use it from the other methods.

pkg/tool/kubectl.go
```diff
@@ -44,8 +47,65 @@ func (k Kubectl) DeleteNamespace(namespace string) {
 	time.Sleep(3 * time.Second)
 
 	if err := k.exec.RunProcess("kubectl", "get", "namespace", namespace); err != nil {
```
unguiculus: Just thinking: Would it make sense to have a receiver method like `RunWithProxy` here which you could pass a function as argument?

pkg/exec/exec.go
cpanato: makes sense. I am very bad at naming things. my daughter is called `girl` 😄

pkg/tool/kubectl.go
cpanato: sounds good, will refactor

cpanato: i did, not sure if this is good, can you PTAL @unguiculus

pkg/tool/kubectl.go
```diff
@@ -54,66 +43,91 @@ func (k Kubectl) DeleteNamespace(namespace string) {
 		fmt.Println("Error deleting pods:", err)
 	}
 
-	time.Sleep(3 * time.Second)
+	fmt.Println("Force-deleting pvcs...")
```
cpanato: if we delete the pvc/pvs/pods after we tried to delete the namespace, did we not fall in the orphan case? that we cannot remove using kubectl?

unguiculus: Deleting the Helm release should already do this. Then, deleting the namespace should also delete all resources within it. If some resources are still stuck terminating after that, we force-delete them. Then we wait again a little while to see if the namespace terminates before we remove its finalizers and force-delete it.

scottrigby: I had this same question. Sounds like a sane strategy 👍

PR #114, opened 2019-02-19
Break docker build if ct is not executable

### LINKED ISSUE #113: Latest Docker image is broken
Running `latest` Docker image produces an error:
```
-> docker run -v $(pwd)/charts:/charts quay.io/helmpack/chart-testing:latest /usr/local/bin/ct lint --all --chart-dirs /charts --debug --validate-maintainers=false
docker: Error response from daemon: OCI runtime create failed: container_linux.go:344: starting container process caused "exec: \"/usr/local/bin/ct\": permission denied": unknown.
ERRO[0001] error waiting for container: context canceled 
```

While the same works fine on `v.2.2.0` tag:
```
docker run -v $(pwd)/charts:/charts quay.io/helmpack/chart-testing:v2.2.0 /usr/local/bin/ct lint --all --chart-dirs /charts --debug --validate-maintainers=false
```

jlegrone:

I think what happened here is that `latest` was built with `docker build .` rather than `build.sh`, and pushed to quay.io. Previously this would have failed when `ct` wasn't found in the build context, but as of a1528a422bebab52029f5014357f2b2a80b7a321 `ct` exists as a directory and is copied without error.

This PR adds a RUN step to ensure that `ct` can execute, causing `docker build .` to fail.

**Which issue this PR fixes**

Prevents #113 (close that issue by building with `build.sh` and pushing again)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,3 +35,5 @@ RUN curl -LO "https://kubernetes-helm.storage.googleapis.com/helm-$HELM_VERSION-
 COPY ./etc/chart_schema.yaml /etc/ct/chart_schema.yaml
 COPY ./etc/lintconf.yaml /etc/ct/lintconf.yaml
 COPY ct /usr/local/bin/ct
+# Ensure that the binary is available on path and is executable
+RUN ct --help


.goreleaser.yml
```diff
@@ -46,5 +46,4 @@ dockers:
       - --label=org.label-schema.description='ct - The chart testing tool'
       - --label=org.label-schema.vendor=Helm
     extra_files:
-      - etc/chart_schema.yaml
-      - etc/lintconf.yaml
+      - etc
```
jlegrone: I'm getting the following error when running `build.sh` locally:

```
COPY failed: stat /var/lib/docker/tmp/docker-builder595575462/etc/chart_schema.yaml: no such file or directory
```

This fixes that, but I'm not sure why. 😦

jlegrone:
cc @unguiculus @cpanato

.goreleaser.yml
unguiculus: Hm, I can run it successfully and so can CircleCI. Not sure why it fails for you.

cpanato: on my mac it failed as well

jlegrone: I'm also on a mac, goreleaser installed via homebrew. Looks like my version is a little old though:

```
0.93.2, commit ce267acabf3205e889ead098a589d905bef32d01, built at 2018-11-12T12:08:25Z
```

I'll try installing a more recent version.

jlegrone: Yeah, no longer a problem with binary installed from https://github.com/goreleaser/goreleaser/releases/tag/v0.101.0.

Should we keep this fix for homebrew users anyway?

PR #117, opened 2019-02-22
Set kind node to 1 worker

rimusz:

Set kind node to 1 worker, as `host-path` storage provider doesn't work in multi-node


unguiculus:
LGTM

PR #119, opened 2019-02-28
add deletion of pv and pvcs

cpanato:
Before we start the namespace cleanup we try to delete any remaining PV/PVC(s) in the namespace.

### DIFF
--- a/pkg/tool/kubectl.go
+++ b/pkg/tool/kubectl.go
@@ -24,12 +24,12 @@ func NewKubectl(exec exec.ProcessExecutor) Kubectl {
 // namespace and, eventually, the namespace itself are force-deleted.
 func (k Kubectl) DeleteNamespace(namespace string) {
 
-	fmt.Println("deleting pvcs...")
+	fmt.Println("Deleting pvcs...")
 	if err := k.exec.RunProcess("kubectl", "delete", "pvc", "--namespace", namespace, "--all"); err != nil {
 		fmt.Println("Error deleting pvc(s):", err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/tool/kubectl.go
```diff
@@ -23,6 +23,17 @@ func NewKubectl(exec exec.ProcessExecutor) Kubectl {
 // DeleteNamespace deletes the specified namespace. If the namespace does not terminate within 90s, pods running in the
 // namespace and, eventually, the namespace itself are force-deleted.
 func (k Kubectl) DeleteNamespace(namespace string) {
+
+	fmt.Println("deleting pvcs...")
```
unguiculus: Match existing style: 

```suggestion
	fmt.Println("Deleting pvcs...")
```

pkg/tool/kubectl.go
```diff
@@ -23,6 +23,17 @@ func NewKubectl(exec exec.ProcessExecutor) Kubectl {
...
+	if err := k.exec.RunProcess("kubectl", "delete", "pvc", "--namespace", namespace, "--all"); err != nil {
+		fmt.Println("Error deleting pvc(s):", err)
+	}
+
+	fmt.Println("deleting pvs...")
```
unguiculus: Match existing style: 

```suggestion
	fmt.Println("Deleting pvs...")
```

cpanato: done

pkg/tool/kubectl.go
cpanato: done

PR #120, opened 2019-03-01
Add Helm/Kubernetes integration tests

jlegrone:

This PR adds tests to verify that helm upgrade and install testing works against a real kubernetes cluster, and runs these tests in CI.

**Which issue this PR fixes**

For review purposes only at the moment; this should be a fast follower to #103.


ct/cmd/install.go
```diff
@@ -31,13 +31,17 @@ func newInstallCmd() *cobra.Command {
 		Use:   "install",
 		Short: "Install and test a chart",
 		Long: heredoc.Doc(`
-			Run 'helm install' and ' helm test' on
+			Run 'helm install', 'helm test', and optionally 'helm upgrade' on
```
scottrigby: Do you want to pull ee32efab79ca923f7baac9e2ed7ac7772bc33ea0 out of this PR now that this is split from #103? It may make this easier to review.

jlegrone: Hmm... I suppose I could point this PR to my own branch, but that's on my fork so I don't think that'll work. Otherwise I don't think there's a way to prevent the diff from #103 showing while also keeping the new tests passing.

jlegrone: Most everything is in https://github.com/helm/chart-testing/pull/120/commits/6c91a169655f6d13fedd9db5f9101758645c231c (with a small change to use recently updated example hostpath configuration in https://github.com/helm/chart-testing/pull/120/commits/bcbb4a50b8fe69b6b3b4cad94c12578d4e242ba5). 😄

scottrigby: Never mind, I now see this depends on #103. You noted it in the initial description 😄 Perfect

e2e-kind.sh
```diff
@@ -0,0 +1,73 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly KIND_VERSION=0.1.0
+readonly CLUSTER_NAME=chart-testing
+readonly K8S_VERSION=v1.13.2
+
+create_kind_cluster() {
+    kind create cluster --name "$CLUSTER_NAME" --config examples/kind/test/kind-config.yaml --image "kindest/node:$K8S_VERSION" --wait 10s
```
unguiculus: I would not use the example config for kind. It's unnecessary as we don't need a multi-mode cluster. I wouldn't rely on examples anyways. They may change independently and break things.

```suggestion
    kind create cluster --name "$CLUSTER_NAME" --image "kindest/node:$K8S_VERSION" --wait 10s
```

.circleci/config.yml
```diff
@@ -10,6 +10,29 @@ jobs:
           command: |
             shellcheck -x build.sh
             shellcheck -x tag.sh
```
unguiculus:
```suggestion
            shellcheck -x tag.sh
            shellcheck -x e2e-kind.sh
```

e2e-kind.sh
jlegrone: > They may change independently and break things.

That was why I went with the example config, I wanted to make sure we provide working examples. 😄 I'm happy to remove this though.

.circleci/config.yml
```diff
@@ -10,6 +10,30 @@ jobs:
...
+      - run:
+          name: Install tools
+          command: |
+            mkdir /home/circleci/.go_workspace/bin
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+            chmod +x ./kubectl
+            sudo mv ./kubectl /usr/local/bin/kubectl
+            curl https://raw.githubusercontent.com/helm/helm/master/scripts/get | bash
+            curl -sSLo kind "https://github.com/kubernetes-sigs/kind/releases/download/0.1.0/kind-linux-amd64"
```
unguiculus: kind v0.2.0 was released two days ago. Maybe we should update.

rimusz: +1

rimusz:
kind `v0.2.0` changed something in networking, so it will not work with the current setup

PR #123, opened 2019-03-05
add require when testing the install charts

cpanato:
I'm using this example with circleci as well and notice a small change we need to do in order to run the tests by adding the parameter `no_output_timeout`, this make the circle ci wait, because the install chart can take some time without any output and in CCi if no output happen in a short time it will kill the build.

Also added a requirement in the install chart to have the lints all passing before install

### DIFF
--- a/examples/kind/.circleci/config.yml
+++ b/examples/kind/.circleci/config.yml
@@ -22,6 +22,7 @@ jobs:
       - checkout
       - run:
           command: test/e2e-kind.sh
+          no_output_timeout: 1h
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


examples/kind/.circleci/config.yml
```diff
@@ -22,11 +22,15 @@ jobs:
       - checkout
       - run:
           command: test/e2e-kind.sh
+          no_output_timeout: 1h
```
unguiculus: I noticed the issue, too, recently. The problem was that CircleCI has a default timeout of 10 minutes which is the same as the Helm timeout we use on charts CI. 

However, chart-testing doesn't set a timeout, so Helm's default of 300 sec applies which doesn't cause this issue. Thus, I don't think we should prematurely change CircleCI timeouts in the example. If a users changes timeouts they should not how to deal with them on CircleCI.

cpanato: makes sense. closing

PR #129, opened 2019-03-20
Switch to go modules and use GoCenter as GOPROXY to speed up go builds

rimusz:

Switch to go modules and use https://gocenter.io as `GOPROXY`.

Using https://gocenter.io as `GOPROXY` will speed modules download and will ensure that modules referenced in `go.mod` are always the same, doesn't matter if some modules get deleted for GitHub and etc, they got cached for live at gocenter.io :-)


rimusz:
cc @unguiculus

jlegrone:
@rimusz I'm not familiar with gocenter -- would this have any impact on local development setup or installing with `go get`?

rimusz:
gocenter.io is goproxy, for local development it is better to set `GO111MODULE=on` and `GOPROXY=https://gocenter.io' and use go v1.11 or v1.12 so you get downloaded exact same modules as in CI

jlegrone:
Actually what's the benefit of using GOPROXY if we continue to vendor our dependencies?

I am a fan of switching to modules btw, although hopefully there will be support for locally serving godoc outside of GOPATH soon.

rimusz:
the bigest plus having gocenter behind the goproxy, is that you always get the same modules you ask for

unguiculus:
The PR title is misleading. It doesn't say anything about switching to modules, which I would support.

I'm hesitant to switch to GoCenter, though. First, I don't think we should promote JFrog stuff here. Second, my experiences with JFrog's JCenter (similar thing, but for Java) are somewhat mixed. I can second this opinion: https://twitter.com/JakeWharton/status/1073102730443526144.

rimusz:
modules are taken from `Gopkg.` files and written to `mod.` files.
Updated about switching to modules.
Gocenter is a free service for Go community, it makes easier to adopt go modules :-)

scottrigby:
Gocenter seems like an interesting project! However for chart-testing I don't think the build time is a pain point yet, and a lot depends on chart-testing to run smoothly – I'm hesitant to add an external tooling dependency at this point.

Whether we switch back to go modules should be a separate question. On that question, early in switching from a bash project to Go, chart-tooling ran into a dependency conflict using modules. I'm not opposed to trying it again, just noting that.

rimusz:
ok, closing the PR

PR #132, opened 2019-03-21
Add option to skip upgrade testing of deleted/renamed values files

### LINKED ISSUE #130: Add flag to disable upgrade tests for renamed or removed values files
Right now switching on the `--upgrade` flag can lead to cases where it's not possible to get `ct` to succeed, based on the contents of a previous revision.

The upgrade flag uses the previous revision's ci values files instead of the current revision's values files, and thus if the previous revision has a ci values file that does not support upgrades  ([example](https://github.com/helm/charts/pull/12384)), it becomes impossible to get `ct` to pass unless you bump the version in `Chart.yaml` a major version.

One approach to this problem is to add a flag that will tell `ct` _not_ to pay attention to values files from the previous revision that have been removed or renamed in the current revision. This will allow simply deleting bad values files at the current revision to get builds passing.

A downside to this would be that some legitimate failure modes could be missed during renames or that folks might be encouraged to simply delete test files when there could be a good reason they're failing.

jlegrone:

See rationale in #130.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*:

Fixes #130 


pkg/chart/chart.go
```diff
@@ -175,6 +175,17 @@ func (c *Chart) ValuesFilePathsForCI() []string {
 	return c.ciValuesPaths
 }
 
+// HasCIValuesFile checks whether a given CI values file is present.
+func (c *Chart) HasCIValuesFile(filePath string) bool {
+	fileName := path.Base(filePath)
```
unguiculus: We should probably use `filepath.Base` instead.

```suggestion
	fileName := filepath.Base(filePath)
```

pkg/chart/chart.go
```diff
@@ -175,6 +175,17 @@ func (c *Chart) ValuesFilePathsForCI() []string {
...
+	for _, file := range c.ValuesFilePathsForCI() {
+		if fileName == path.Base(file) {
```
unguiculus:
```suggestion
		if fileName == filepath.Base(file) {
```

pkg/chart/chart.go
jlegrone: Good call.

Issue #60, opened 2018-11-17
Exit gracefully (run cleanup tasks) when relevant unix signals are received


jlegrone:
When an interrupt signal is received, for example, any releases or namespaces that are pending deletion should be deleted. This would be great for local usage of `ct install` when the user may wish to exit the program as soon as it's clear that a deployment has failed.

PR #133, opened 2019-03-22
Add git version documentation, rename worktree methods, and handle existing worktree

jlegrone:

- 8709dde729f4648fb2f08ce96e73ac4d3d117193: Renaming `*WorkingTree` to `*WorkTree` is more consistent with the underlying git commands.
- 37e686641554982be48f3222d697f983df3205e6: Now that we use `git worktree`, we should document which git versions work with `ct`.
- d0d57443dd0f2f2402c85986e49754b85c249858: There is an edge case where `ct` will create a new worktree and then not clean it up if execution is cancelled. This causes `ct` to fail on the next run because `git worktree add` errors if the worktree path already exists. Creating a temp directory for each run fixes that and also should prevent the accidental overwrite of a user-managed git worktree. We should handle the cleanup of the temp directory as part of #60 in the future.

**Which issue this PR fixes**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ It is recommended to use the provided Docker image which can be [found on Quay](
 It comes with all necessary tools installed.
 
 * [helm](http://helm.sh)
+* [git](https://git-scm.com) (2.17.0 or later)
 * [yamllint](https://github.com/adrienverge/yamllint)
 * [yamale](https://github.com/23andMe/Yamale)
 * [kubectl](https://kubernetes.io/docs/reference/kubectl/overview/)
--- a/README.md
+++ b/README.md
@@ -15,11 +15,11 @@ It automatically detects charts changed against the target branch.
 It is recommended to use the provided Docker image which can be [found on Quay](https://quay.io/helmpack/chart-testing/).
 It comes with all necessary tools installed.
 
-* [helm](http://helm.sh)
-* [git](https://git-scm.com) (2.17.0 or later)
-* [yamllint](https://github.com/adrienverge/yamllint)
-* [yamale](https://github.com/23andMe/Yamale)
-* [kubectl](https://kubernetes.io/docs/reference/kubectl/overview/)
+* [Helm](http://helm.sh)
+* [Git](https://git-scm.com) (2.17.0 or later)
+* [Yamllint](https://github.com/adrienverge/yamllint)
+* [Yamale](https://github.com/23andMe/Yamale)
+* [Kubectl](https://kubernetes.io/docs/reference/kubectl/overview/)
 
 ### Binary Distribution
 
--- a/README.md
+++ b/README.md
@@ -15,11 +15,10 @@ It automatically detects charts changed against the target branch.
 It is recommended to use the provided Docker image which can be [found on Quay](https://quay.io/helmpack/chart-testing/).
 It comes with all necessary tools installed.
 
-* Helm (http://helm.sh)
-* yamllint (https://github.com/adrienverge/yamllint)
-* yamale (https://github.com/23andMe/Yamale)
-* kubectl (https://kubernetes.io/docs/reference/kubectl/overview/)
-* Tooling for your cluster
+* [helm](http://helm.sh)
+* [yamllint](https://github.com/adrienverge/yamllint)
+* [yamale](https://github.com/23andMe/Yamale)
+* [kubectl](https://kubernetes.io/docs/reference/kubectl/overview/)
 
 ### Binary Distribution
 
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -35,9 +35,9 @@ const maxNameLength = 63
 //
 // Show returns the contents of file on the specified remote/branch.
 //
-// AddWorkTree checks out the contents of the repository at a commit ref into the specified path.
+// AddWorktree checks out the contents of the repository at a commit ref into the specified path.
 //
-// RemoveWorkTree removes the working tree at the specified path.
+// RemoveWorktree removes the working tree at the specified path.
 //
 // MergeBase returns the SHA1 of the merge base of commit1 and commit2.
 //
@@ -50,8 +50,8 @@ const maxNameLength = 63
 type Git interface {
 	FileExistsOnBranch(file string, remote string, branch string) bool
 	Show(file string, remote string, branch string) (string, error)
-	AddWorkTree(path string, ref string) error
-	RemoveWorkTree(path string) error
+	AddWorktree(path string, ref string) error
+	RemoveWorktree(path string) error
 	MergeBase(commit1 string, commit2 string) (string, error)
 	ListChangedFilesInDirs(commit string, dirs ...string) ([]string, error)
 	GetUrlForRemote(remote string) (string, error)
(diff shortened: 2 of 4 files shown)


pkg/chart/chart.go
```diff
@@ -314,8 +314,13 @@ func (t *Testing) processCharts(action func(chart *Chart) TestResult) ([]TestRes
 		if err != nil {
 			return results, errors.Wrap(err, "Error identifying merge base")
 		}
-		t.git.AddWorkingTree(ctPreviousRevisionTree, mergeBase)
-		defer t.git.RemoveWorkingTree(ctPreviousRevisionTree)
+		// Remove worktree first (it may not have been cleaned up correctly from the last test run)
+		// TODO: This may no longer be needed once https://github.com/helm/chart-testing/issues/60
+		//       is addressed.
+		t.git.RemoveWorkTree(ctPreviousRevisionTree)
```
jlegrone: There's a (very small) chance that we could be deleting an actively used worktree here. Currently our worktree is created at the path `ct_previous_revision`, and this seems fine to me.

To be _extra_ safe we could set the worktree path to something even more clearly owned by helm, like `helm.sh/ct_previous_revision`. Thoughts?

README.md
```diff
@@ -15,11 +15,11 @@ It automatically detects charts changed against the target branch.
 It is recommended to use the provided Docker image which can be [found on Quay](https://quay.io/helmpack/chart-testing/).
 It comes with all necessary tools installed.
 
-* Helm (http://helm.sh)
-* yamllint (https://github.com/adrienverge/yamllint)
-* yamale (https://github.com/23andMe/Yamale)
-* kubectl (https://kubernetes.io/docs/reference/kubectl/overview/)
-* Tooling for your cluster
```
jlegrone: Wasn't sure what this meant exactly. Happy to add this back if we can get a little more specific.

pkg/chart/chart.go
unguiculus: What about creating a temp dir? Something like this:

`ioutil.TempFile("helm.sh", "ct_previous_revision")`

README.md
unguiculus: I think this can be removed.

pkg/chart/chart.go
jlegrone: @unguiculus it does seem like this would be a great approach, but we'd probably need to clean up the log formatting. I tried using `ioutil.TempDir` and on mac it makes the logs much less user-friendly:

```
Installing chart '/var/folders/09/t_s1mmqd3ls74fd5b6h4hg0h0000gp/T/ct_previous_revision731558328/stable/redis' with values file '/var/folders/09/t_s1mmqd3ls74fd5b6h4hg0h0000gp/T/ct_previous_revision731558328/stable/redis/ci/default-values.yaml'...
```

Maybe we should go with the current approach for now? I think it's also potentially nice to be able to debug with the git worktree at a relative path. With the `helm.sh` prefix, logs look like this:

```
Installing chart 'helm.sh/ct_previous_revision/stable/redis' with values file 'helm.sh/ct_previous_revision/stable/redis/ci/default-values.yaml'...
```

unguiculus: Of course, it's `ioutil.TempDir`, not `ioutil.TempFile` as I had suggested. If you call it with the `dir` parameter, it creates the temp dir in this directory. Only if you pass an empty string, it delegates to `os.TempDir` to compute the directory.

jlegrone: Okay so I did a little more experimentation. In order for `ioutil.TempDir` to work inside `helm.sh`, `helm.sh` has to already exist. We could create it first but that's kind of a mess since we'd need to do some bookkeeping to make sure we never delete `helm.sh` if it already existed or its contents changed outside of `ct`.

Second approach: Just generate a worktree path at the top level of the repo. This is much simpler, but the downside is that it's a tiny bit ugly:

```
Installing chart 'ct_previous_revision223934316/stable/redis' with values file 'ct_previous_revision223934316/stable/redis/ci/default-values.yaml'...
```

jlegrone: @unguiculus thoughts on the top-level tempdir vs `helm.sh/ct_previous_revision`?

scottrigby: What about something simple like this? 
> If path is already a directory, MkdirAll does nothing and returns nil.
```diff
	if err := os.MkdirAll("helm.sh", 0755); err != nil {
		fmt.Println(errors.Wrapf(err, "Error creating worktree directory for previous revision of chart '%s'\n", chart))
	}
```

unguiculus: I'm fine with top-level or @scottrigby's solution.

scottrigby: Same

jlegrone: Cool, I'll push the top-level changes up then. `os.MkdirAll` was what I was alluding to in my [above comment](https://github.com/helm/chart-testing/pull/133#discussion_r268778090) -- we run into the same problems if we want to avoid creating side effects as a result of running `ct`.

pkg/chart/chart_test.go
```diff
@@ -31,17 +29,11 @@ import (
...
-	return err == nil
+	return true
 }
 
 func (g fakeGit) Show(file string, remote string, branch string) (string, error) {
-	b, err := ioutil.ReadFile(computePreviousRevisionPath(file))
-	if err != nil {
-		return "", err
-	}
-	return string(b), nil
```
jlegrone: This stuff was no longer necessary after https://github.com/helm/chart-testing/pull/120/commits/9616010a874d8b6bf698a2cb6e36381848119044.

README.md
```diff
@@ -15,11 +15,11 @@ It automatically detects charts changed against the target branch.
...
+* [helm](http://helm.sh)
```
unguiculus:
```suggestion
* [Helm](http://helm.sh)
```

README.md
```diff
@@ -15,11 +15,11 @@ It automatically detects charts changed against the target branch.
...
+* [git](https://git-scm.com) (2.17.0 or later)
```
unguiculus:
```suggestion
* [Git](https://git-scm.com) (2.17.0 or later)
```

pkg/chart/chart.go
```diff
@@ -35,9 +35,9 @@ const maxNameLength = 63
 //
 // Show returns the contents of file on the specified remote/branch.
 //
-// AddWorkingTree checks out the contents of the repository at a commit ref into the specified path.
+// AddWorkTree checks out the contents of the repository at a commit ref into the specified path.
```
unguiculus: I'd rename as follows to be consistent with Git:
WorkTree -> Worktree

pkg/chart/chart.go
```diff
@@ -253,12 +254,10 @@ func NewTesting(config config.Configuration) Testing {
...
 }
 
-const ctPreviousRevisionTree = "ct_previous_revision"
-
 // computePreviousRevisionPath converts any file or directory path to the same path in the
 // previous revision's working tree.
-func computePreviousRevisionPath(fileOrDirPath string) string {
-	return filepath.Join(ctPreviousRevisionTree, fileOrDirPath)
+func (t *Testing) computePreviousRevisionPath(fileOrDirPath string) string {
+	return filepath.Join(t.previousRevisionWorkTree, fileOrDirPath)
```
unguiculus:
```suggestion
	return filepath.Join(t.previousRevisionWorktree, fileOrDirPath)
```

pkg/chart/chart.go
```diff
@@ -324,11 +323,20 @@ func (t *Testing) processCharts(action func(chart *Chart) TestResult) ([]TestRes
...
 			return results, errors.Wrap(err, "Error identifying merge base")
 		}
-		t.git.AddWorkingTree(ctPreviousRevisionTree, mergeBase)
-		defer t.git.RemoveWorkingTree(ctPreviousRevisionTree)
+		// Add worktree for the target revision
+		worktreePath, err := ioutil.TempDir("./", "ct_previous_revision")
+		if err != nil {
+			return results, errors.Wrap(err, "Could not create previous revision directory")
+		}
+		t.previousRevisionWorkTree = worktreePath
```
unguiculus:
```suggestion
		t.previousRevisionWorktree = worktreePath
```

pkg/chart/chart.go
```diff
@@ -215,14 +215,15 @@ func NewChart(chartPath string) (*Chart, error) {
...
-	chartUtils       ChartUtils
+	config                   config.Configuration
+	helm                     Helm
+	kubectl                  Kubectl
+	git                      Git
+	linter                   Linter
+	accountValidator         AccountValidator
+	directoryLister          DirectoryLister
+	chartUtils               ChartUtils
+	previousRevisionWorkTree string
```
unguiculus:
```suggestion
	previousRevisionWorktree string
```

PR #135, opened 2019-03-26
Update kind to v0.2.0

rimusz:

### DIFF
--- a/examples/docker-for-mac/my_test.sh
+++ b/examples/docker-for-mac/my_test.sh
@@ -21,15 +21,24 @@ set -o pipefail
 readonly IMAGE_TAG=v2.2.0
 readonly IMAGE_REPOSITORY="quay.io/helmpack/chart-testing"
 
-main() {
-    local testcontainer_id
-    testcontainer_id=$(create_testcontainer)
+run_ct_container() {
+    echo 'Running ct container...'
+    docker container run --rm --interactive --detach --name ct \
+        --volume "$(pwd):/workdir" \
+        --workdir /workdir \
+        "$IMAGE_REPOSITORY:$IMAGE_TAG"
+    echo
+}
 
-    # shellcheck disable=SC2064
-    trap "docker container rm --force $testcontainer_id > /dev/null" EXIT
+cleanup() {
+    echo 'Removing ct container...'
+    docker kill ct > /dev/null 2>&1
 
-    configure_kubectl "$testcontainer_id"
-    run_test
+    echo 'Done!'
+}
+
+docker_exec() {
+    docker exec --interactive -e HELM_HOST=127.0.0.1:44134 -e HELM_TILLER_SILENT=true ct "$@"
 }
 
 lookup_apiserver_container_id() {
@@ -42,15 +51,7 @@ get_apiserver_arg() {
     docker container inspect "$container_id" | jq -r ".[].Args[] | capture(\"$arg=(?<arg>.*)\") | .arg"
 }
 
-create_testcontainer() {
-    docker container run --interactive --tty --detach \
-        --volume "$(pwd):/workdir" --workdir /workdir \
-        "$IMAGE_REPOSITORY:$IMAGE_TAG" cat
-}
-
-configure_kubectl() {
-    local testcontainer_id="$1"
-
+connect_to_cluster() {
     local apiserver_id
     apiserver_id=$(lookup_apiserver_container_id)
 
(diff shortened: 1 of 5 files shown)


examples/docker-for-mac/my_test.sh
```diff
@@ -65,17 +66,37 @@ configure_kubectl() {
...
+    echo 'Installing Tiller...'
+    docker_exec kubectl --namespace kube-system create sa tiller
+    docker_exec kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
+    docker_exec helm init --service-account tiller --upgrade --wait
+    echo
+}
+
+install_charts() {
+    echo "Add git remote k8s ${CHARTS_REPO}"
+    git remote add k8s "${CHARTS_REPO}" &> /dev/null || true
```
unguiculus: `CHARTS_REPO` is not defined. Why change this? Maybe we shjould emove the separate remote which is just confusing sand use `origin` which is the default.

examples/docker-for-mac/my_test.sh
```diff
@@ -65,17 +66,37 @@ configure_kubectl() {
...
+    git fetch k8s master
+    echo
+    # shellcheck disable=SC2086
+    docker_exec ct install ${CHART_TESTING_ARGS} --config /workdir/test/ct.yaml
```
unguiculus: `CHART_TESTING_ARGS` is not defined. Why add this? Please remove it again.

examples/docker-for-mac/my_test.sh
```diff
@@ -21,15 +21,24 @@ set -o pipefail
 readonly IMAGE_TAG=v2.2.0
 readonly IMAGE_REPOSITORY="quay.io/helmpack/chart-testing"
 
-main() {
-    local testcontainer_id
-    testcontainer_id=$(create_testcontainer)
+run_ct_container() {
+    echo 'Running ct container...'
+    docker container run --rm --interactive --detach --name ct \
```
unguiculus: You should leave the existing behavior and not use a hard-coded container name. Otherwise you won't be able to do parallel executions.

examples/docker-for-mac/my_test.sh
```diff
@@ -21,15 +21,24 @@ set -o pipefail
...
+        --workdir /workdir \
+        "$IMAGE_REPOSITORY:$IMAGE_TAG"
+    echo
+}
 
-    # shellcheck disable=SC2064
-    trap "docker container rm --force $testcontainer_id > /dev/null" EXIT
+cleanup() {
+    echo 'Removing ct container...'
+    docker kill ct > /dev/null 2>&1
```
unguiculus: This no longer removes the container.

examples/docker-for-mac/my_test.sh
```diff
@@ -65,17 +66,37 @@ configure_kubectl() {
...
+    docker_exec helm init --service-account tiller --upgrade --wait
+    echo
+}
+
+install_charts() {
+    echo "Add git remote k8s ${CHARTS_REPO}"
+    git remote add k8s "${CHARTS_REPO}" &> /dev/null || true
+    git fetch k8s master
+    echo
+    # shellcheck disable=SC2086
```
unguiculus: Why disable shellcheck here?

examples/docker-for-mac/my_test.sh
```diff
@@ -21,15 +21,24 @@ set -o pipefail
...
+    echo 'Removing ct container...'
+    docker kill ct > /dev/null 2>&1
+
+    echo 'Done!'
+}
 
-    configure_kubectl "$testcontainer_id"
-    run_test
+docker_exec() {
+    docker exec --interactive -e HELM_HOST=127.0.0.1:44134 -e HELM_TILLER_SILENT=true ct "$@"
```
unguiculus: Why do you add `HELM_HOST` and `HELM_TILLER_SILENT`. This sounds like tillerless Helm which we don't do here.

scottrigby: If we do want to support people using for testing tillerless helm, we could do that easily with a [bash param substitution default](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion) to false (`-e HELM_TILLER_SILENT=${HELM_TILLER_SILENT:-false}`), and then those who want can `export HELM_TILLER_SILENT=true` before running the script? Either way perhaps this should be a separate PR so we can discuss whether and how to support that option.

unguiculus: After all, it's just an example. We don't have to cater for every possible use case here.

PR #143, opened 2019-04-19
Trim leading hyphens from names

unguiculus:


unguiculus:
Related: https://github.com/helm/charts/pull/12951

PR #144, opened 2019-04-19
Fix kind timeout

unguiculus:


jlegrone:
Should this also be updated in https://github.com/helm/chart-testing/blob/0c36eef46e78f51f20e4cd9233f5f25833495a33/examples/kind/test/e2e-kind.sh?

PR #145, opened 2019-04-23
Update local path provisioner version in the example folder

cpanato:
Update local path prov to use version 0.0.6

### DIFF
--- a/examples/kind/test/e2e-kind.sh
+++ b/examples/kind/test/e2e-kind.sh
@@ -5,9 +5,9 @@ set -o nounset
 set -o pipefail
 
 readonly CT_VERSION=v2.3.3
-readonly KIND_VERSION=0.2.1
+readonly KIND_VERSION=v0.3.0
 readonly CLUSTER_NAME=chart-testing
-readonly K8S_VERSION=v1.13.4
+readonly K8S_VERSION=v1.14.2
 
 run_ct_container() {
     echo 'Running ct container...'
--- a/examples/kind/test/e2e-kind.sh
+++ b/examples/kind/test/e2e-kind.sh
@@ -5,9 +5,9 @@ set -o nounset
 set -o pipefail
 
 readonly CT_VERSION=v2.3.3
-readonly KIND_VERSION=v0.3.0
+readonly KIND_VERSION=v0.5.1
 readonly CLUSTER_NAME=chart-testing
-readonly K8S_VERSION=v1.14.2
+readonly K8S_VERSION=v1.15.3
 
 run_ct_container() {
     echo 'Running ct container...'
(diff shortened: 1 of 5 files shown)


examples/kind/test/e2e-kind.sh
```diff
@@ -7,7 +7,7 @@ set -o pipefail
 readonly CT_VERSION=v2.2.0
 readonly KIND_VERSION=0.2.1
 readonly CLUSTER_NAME=chart-testing
-readonly K8S_VERSION=v1.13.4
+readonly K8S_VERSION=v1.14.1
```
jlegrone: Can we make the same change in https://github.com/helm/chart-testing/blob/d9dba63311e5d0524832bf35e9c83993a76af93d/e2e-kind.sh?

cpanato:
now it is released: https://github.com/rancher/local-path-provisioner/releases

I will address the comments above as well

examples/kind/test/local-path-provisioner.yaml
```diff
@@ -60,7 +60,7 @@ spec:
       serviceAccountName: local-path-provisioner-service-account
       containers:
       - name: local-path-provisioner
-        image: rancher/local-path-provisioner:v0.0.5
+        image: rancher/local-path-provisioner:v0.0.6
```
jlegrone: Looks like there's another version out?

```suggestion
        image: rancher/local-path-provisioner:v0.0.7
```

scottrigby: Now https://github.com/rancher/local-path-provisioner/releases/tag/v0.0.8. Moving pretty fast

cpanato: 0.0.9 :) updated

examples/kind/test/e2e-kind.sh
cpanato: done

cpanato: done

cpanato:
@scottrigby @jlegrone @unguiculus PTAL

examples/kind/test/kind-config.yaml
```diff
@@ -1,5 +1,5 @@
 kind: Cluster
-apiVersion: kind.sigs.k8s.io/v1alpha3
+apiVersion: kind.sigs.k8s.io/v1alpha2
```
unguiculus: Why did you change this?

cpanato: I will check because i did locally to use the new kind and the charts installation dod not work

unguiculus: I have experienced an issue with kind 0.3.0. Might be due to this. We should skip 0.3.0 and go to 0.4.0 directly.

cpanato: setting to 0.5.1
sorry for the long delay.... :(

PR #152, opened 2019-04-24
Fix log messages

unguiculus:

### DIFF
--- a/pkg/tool/git.go
+++ b/pkg/tool/git.go
@@ -72,5 +72,6 @@ func (g Git) GetUrlForRemote(remote string) (string, error) {
 }
 
 func (g Git) ValidateRepository() error {
-	return g.exec.RunProcess("git", "rev-parse", "--is-inside-work-tree")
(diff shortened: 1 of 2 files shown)


pkg/tool/git.go
```diff
@@ -72,5 +72,6 @@ func (g Git) GetUrlForRemote(remote string) (string, error) {
 }
 
 func (g Git) ValidateRepository() error {
-	return g.exec.RunProcess("git", "rev-parse", "--is-inside-work-tree")
+	_, err := g.exec.RunProcessAndCaptureOutput("git", "rev-parse", "--is-inside-work-tree")
+	return err
```
jlegrone: Is this functionally any different from what was implemented before?

unguiculus: The difference is that it doesn't log anything. The output of `git rev-parse --is-inside-work-tree` is `true`, which was logged but didn't make much sense as such.

PR #155, opened 2019-05-02
Print events on cleanup

unguiculus:
Helps identifying issues when Helm timeout because pods cannot be created due to errors.

Affected PR: https://github.com/helm/charts/pull/13444

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -99,6 +99,8 @@ type Helm interface {
 //
 // GetPods gets pods for the given args
 //
+// GetEvents prints all events for namespace
+//
 // DescribePod prints the pod's description
 //
 // Logs prints the logs of container
@@ -111,6 +113,7 @@ type Kubectl interface {
 	WaitForDeployments(namespace string, selector string) error
 	GetPodsforDeployment(namespace string, deployment string) ([]string, error)
 	GetPods(args ...string) ([]string, error)
+	GetEvents(namespace string) error
 	DescribePod(namespace string, pod string) error
 	Logs(namespace string, pod string, container string) error
 	GetInitContainers(namespace string, pod string) ([]string, error)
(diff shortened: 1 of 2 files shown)


pkg/chart/chart.go
```diff
@@ -812,7 +815,13 @@ func (t *Testing) ValidateMaintainers(chart *Chart) error {
 	return nil
 }
 
-func (t *Testing) PrintPodDetailsAndLogs(namespace string, selector string) {
+func (t *Testing) PrintEventsPodDetailsAndLogs(namespace string, selector string) {
+	util.PrintDelimiterLine("=")
+
+	printDetails(namespace, "Events of namespace", ".", func(item string) error {
+		return t.kubectl.GetEvents(namespace)
```
jlegrone: Is it possible to use a label selector for events?

unguiculus: Yes, that's possible. I guess that might be better if someone uses a pre-defined namespace

unguiculus: Actually, that won't quite work because the events won't have the labels of the chart resources.

unguiculus: You could do something like this:

`kubectl get events --field-selector 'involvedObject.name=foo'`

But that could be tricky. Depending on resource naming we might miss something.

scottrigby:
I like this 👍

pkg/chart/chart.go
jlegrone: 😞that's too bad... as long as tons of existing events doesn't bog down performance or place unnecessary load on the api server this seems okay for either case though.

pkg/tool/kubectl.go
```diff
@@ -184,6 +184,10 @@ func (k Kubectl) GetPods(args ...string) ([]string, error) {
 	return strings.Fields(pods), nil
 }
 
+func (k Kubectl) GetEvents(namespace string) error {
+	return k.exec.RunProcess("kubectl", "get", "events", "--namespace", namespace)
```
jlegrone: Adding `-o wide` also includes a `COUNT` column which could be useful.

```suggestion
	return k.exec.RunProcess("kubectl", "get", "events", "--output", "wide", "--namespace", namespace)
```

pkg/chart/chart.go
unguiculus: I dont see an issue here.

kpucynski:
It seems that this PR broke master build:
```
$ ./build.sh 
# chart-testing/pkg/chart [chart-testing/pkg/chart.test]
pkg/chart/chart.go:252:3: cannot use tool.NewKubectl(procExec) (type tool.Kubectl) as type Kubectl in field value:
	tool.Kubectl does not implement Kubectl (missing GetEvents method)
?   	chart-testing/ct	[no test files]
?   	chart-testing/ct/cmd	[no test files]
FAIL	chart-testing/pkg/chart [build failed]
ok  	chart-testing/pkg/config	0.002s
?   	chart-testing/pkg/exec	[no test files]
ok  	chart-testing/pkg/tool	0.011s
ok  	chart-testing/pkg/util	0.006s
FAIL
$ go version go1.13.1 linux/amd64
```

Issue #64, opened 2018-12-04
Add option to git diff between last 2 local commits


monotek:
**Is this a request for help?**:

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
FEATURE REQUEST

**Version of Helm and Kubernetes**:
helm version: 2.11
kubectl version: 1.11
ct version: 2.0.1

**What happened**:
Please add an option to git diff between the last 2 local commits instead between local and remote repo.

I'd like to use ct on a local git repo which was cloned from a remote and therefore hase same commit locally as on the remote. Unfortunately the change detection don't works in this case and i only can use --all to get my charts tested.

**What you expected to happen**:
An option is available to enable change detection to use a diff to HEAD~1 instead to the remote.

**How to reproduce it** (as minimally and precisely as possible):
* Checkout your repo from remote to a local directory
* run ct lint locally
* "No chart changes detected."

unguiculus:
Coud your explain your workflow? Generally speaking, this tool is meant to test changes against a target branch before they are merged. In your case, the chart already seems to be on the target branch. Why is this? Can you elaborate?

monotek:
I've pushed directly to master without using a pr but still wanted the pipline to do the tests...

unguiculus:
I assume this is the repo: https://github.com/monotek/helm-charts

Give this is a repo that's supposed to be added to the hub, I'd say it would make sense to test charts before they make it to master.

monotek:
Yes, of course. The repo was more or less a test to copy the old helm/stable repo workflow without the need to have own infrastructure. At the end i wanted to have the repo on Github pages (https://monotek.github.io/charts/index.yaml) and the upload to that repo should be triggered after succesful test in https://github.com/monotek/helm-charts.

auhlig:
I agree that it would be quite useful to not only diff to `HEAD` but to a configurable commit passed via a flag:
Tested and valid helm charts live on the master branch and changes can only be done by opening a PR and ensuring the helm chart tests pass. 
After merging the PR I want to `git diff ..` with the previous commit `HEAD~1` to be able to detect the changes, test again and finally push the chart to a helm repository.

kyschouv:
This would be useful. We were able to work around it by setting `target-branch` to `master^` if we detected that `HEAD` and `origin/master` (our remote and target branch) were the same commit hash (and that there were no diffs in the files).

PR #159, opened 2019-07-01
allow setting git-ref for merge-base

auhlig:
Allows setting the `commit` used to get the git merge base. Closes https://github.com/helm/chart-testing/issues/64

### DIFF
--- a/ct/cmd/root.go
+++ b/ct/cmd/root.go
@@ -63,6 +63,7 @@ func addCommonFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&cfgFile, "config", "", "Config file")
 	flags.String("remote", "origin", "The name of the Git remote used to identify changed charts")
 	flags.String("target-branch", "master", "The name of the target branch used to identify changed charts")
+	flags.String("commit", "HEAD", "The commit used to identify changed charts")
 	flags.StringSlice("chart-dirs", []string{"charts"}, heredoc.Doc(`
 		Directories containing Helm charts. May be specified multiple times
 		or separate values with commas`))
--- a/ct/cmd/root.go
+++ b/ct/cmd/root.go
@@ -71,7 +71,7 @@ func addCommonFlags(flags *pflag.FlagSet) {
 		The git reference used to identify changed charts`))
 	// Deprecated: Please use `--since` instead.
 	flags.String("remote", "origin", heredoc.Doc(`
-		Deprecated: Use --since=$git-ref instead. The name of the Git remote used to identify changed charts`))
+		The name of the Git remote used to identify changed charts`))
 	// Deprecated: Please use `--since` instead.
 	flags.String("target-branch", "master", heredoc.Doc(`
 		Deprecated: Use --since=$git-ref instead. The name of the target branch used to identify changed charts`))
--- a/ct/cmd/root.go
+++ b/ct/cmd/root.go
@@ -61,15 +61,20 @@ func Execute() {
 
 func addCommonFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&cfgFile, "config", "", "Config file")
-	flags.String("remote", "origin", "The name of the Git remote used to identify changed charts")
-	flags.String("target-branch", "master", "The name of the target branch used to identify changed charts")
-	flags.String("commit", "HEAD", "The commit used to identify changed charts")
 	flags.StringSlice("chart-dirs", []string{"charts"}, heredoc.Doc(`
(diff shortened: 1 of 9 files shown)


ct/cmd/root.go
```diff
@@ -63,6 +63,7 @@ func addCommonFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&cfgFile, "config", "", "Config file")
 	flags.String("remote", "origin", "The name of the Git remote used to identify changed charts")
 	flags.String("target-branch", "master", "The name of the target branch used to identify changed charts")
+	flags.String("commit", "HEAD", "The commit used to identify changed charts")
```
jlegrone: > from https://github.com/helm/chart-testing/issues/64#issuecomment-507276419: After merging the PR I want to git diff .. with the previous commit `HEAD~1` to be able to detect the changes, test again and finally push the chart to a helm repository.

This is confusing to me. Why wouldn't `HEAD` work in this case?

If there is a use case for this, it is still possible to change `HEAD` by doing a `git checkout <ref>` before running `ct`.

auhlig:
Hi @jlegrone,
Thanks for taking the time to review this.

Maybe I misunderstood #64. But +1 for the `--since` flag and deprecating `--target-branch`. I'll do it right away.

pkg/tool/git.go
```diff
@@ -67,8 +67,8 @@ func (g Git) ListChangedFilesInDirs(commit string, dirs ...string) ([]string, er
 	return strings.Split(changedChartFilesString, "\n"), nil
 }
 
-func (g Git) GetUrlForRemote(remote string) (string, error) {
-	return g.exec.RunProcessAndCaptureOutput("git", "ls-remote", "--get-url", remote)
+func (g Git) GetUrlForRemote(ref string) (string, error) {
+	return g.exec.RunProcessAndCaptureOutput("git", "ls-remote", "--get-url", ref)
```
jlegrone: I don't think this will work... we may need to keep the `--remote` flag after all since it's possible that multiple remotes will have the same ref.

auhlig: Indeed. Will restore `--remote`.

jlegrone: Let's get @unguiculus's input before doing too much more work.

pkg/config/config.go
```diff
@@ -118,8 +121,8 @@ func LoadConfiguration(cfgFile string, cmd *cobra.Command, printConfig bool) (*C
 	// Disable upgrade (this does some expensive dependency building on previous revisions)
 	// when neither "install" nor "lint-and-install" have not been specified.
 	cfg.Upgrade = isInstall && cfg.Upgrade
-	if (cfg.TargetBranch == "" || cfg.Remote == "") && cfg.Upgrade {
-		return nil, errors.New("specifying '--upgrade=true' without '--target-branch' or '--remote', is not allowed")
+	if (cfg.Since == "") && cfg.Upgrade {
+		return nil, errors.New("specifying '--upgrade=true' without '--since' is not allowed")
```
unguiculus: That's a breaking change. We need to allow the existing behavior.

pkg/tool/git.go
unguiculus: I will have to think this through. A `--since` flag probably makes sense, but we need to keep `--remote` for now.

scottrigby:
Any changes to `--remote` or `--target-branch` will have to be rethought in v3. We have just created a new `dev-v2` branch for v2 maintenance. See https://github.com/helm/chart-testing/issues/182

jeff-knurek:
We have a use case that requires this. @auhlig are you still working on it? Seems like there's a breaking change in `pkg/config/config.go` that needs to be addressed. 

@scottrigby / @unguiculus  / @jlegrone is there anything else that's needed for this PR to be added to `v2`?

scottrigby:
@jeff-knurek would you please define the use case? @auhlig I had meant to ask you the same. It may help determine how (or whether or not) to implement this.

auhlig:
Assuming a chart exists on the master branch, which is forked and updated or being worked in in another branch. The change would help to detect the changes compared to origin/master:HEAD .

Another use case: We allow only PRs to be merged to the master branch if they pass the (chart) tests. We want to determine which chart(s) was/were updated in the merge commit and  subsequently trigger an upload to a chart registry. For that we use the previous commit as reference via `ct list-changed --commit HEAD~1`.

jeff-knurek:
we want the ability to run the equivalent of `git merge-base HEAD^ HEAD` after a PR is squash/merged to master, and have `ct list-changed` list the changed charts since the previous commit

PR #160, opened 2019-07-01
allow include subcharts

auhlig:
Assuming an umbrella helm chart called `kube-monitoring` with multiple subcharts that looks like this
```
└── kube-monitoring
    ├── charts
    │     ├── prometheus
    │     ├── grafana
...
```

If one applies a change to the `prometheus` chart, the `ct list-changed ...` would only show that the umbrella chart `kube-monitoring` changed but not the subchart.

This PR adds an option to also include the subcharts in the output of `ct list-changed ..` in addition to the umbrella chart:
```
kube-monitoring
kube-monitoring/charts/prometheus
```

### DIFF
--- a/ct/cmd/listChanged.go
+++ b/ct/cmd/listChanged.go
@@ -19,10 +19,10 @@ import (
 	"os"
 
 	"github.com/MakeNowJust/heredoc"
-
 	"github.com/helm/chart-testing/pkg/chart"
 	"github.com/helm/chart-testing/pkg/config"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 func newListChangedCmd() *cobra.Command {
@@ -38,6 +38,7 @@ func newListChangedCmd() *cobra.Command {
 
 	flags := cmd.Flags()
 	addCommonFlags(flags)
+	addListChangedFlags(flags)
 	return cmd
 }
 
@@ -57,4 +58,8 @@ func listChanged(cmd *cobra.Command, args []string) {
 	for _, dir := range chartDirs {
 		fmt.Println(dir)
 	}
+}
+
+func addListChangedFlags(flags *pflag.FlagSet) {
+	flags.Bool("include-subcharts", false, heredoc.Doc("Whether to also include subcharts. (default: false)"))
 }
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -141,7 +141,7 @@ type DirectoryLister interface {
 //
 // LookupChartDir looks up the chart's root directory based on some chart file that has changed
 type ChartUtils interface {
-	LookupChartDir(chartDirs []string, dir string) (string, error)
+	LookupChartDir(chartDirs []string, dir string, isIncludeSubCharts bool) (string, error)
 }
 
 // AccountValidator is the interface that wraps Git account validation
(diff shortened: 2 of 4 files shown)


jlegrone:
This use case is a little bit non-standard -- typically you would declare the `prometheus` chart a dependency via `requirements.yaml` (and `ct` would build `kube-monitoring` at runtime without the need to vendor `prometheus`).

The behavior of only testing the top-level chart is also desirable in this scenario, as long as the prometheus chart is not meant for consumption as a dependency of other charts (if it were it would make more sense to make it a sibling of `kube-monitoring` in version control rather than a child). Running tests only for the ci values provided by the parent chart is more efficient than testing the parent _and_ the ci configuration for each child chart.

auhlig:
Hi @jlegrone,
Thanks for taking the time to review this.

I guess `prometheus` and `grafana` were bad examples but the pattern of having umbrella charts are quite standard. 
Here's a [real example](https://github.com/sapcc/helm-charts/tree/master/system/kube-system).
It's our `kube-system` umbrella chart which contains the stable/nginx-ingress (via requirements.yaml) but also a set of custom charts for our applications.
The structure looks like this:
```
└── kube-system
    ├── charts
    │     ├── disco
    │     ├── vice-president
...
```

As you say: In this case the `ct list-changed ..` would of course be able to detect changes of chart mentioned in the requirements of the `kube-system` umbrella chart.
However, it wouldn't do so for the nested charts `disco`, `vice-president`, etc. - For this we'd need to return the path to the subcharts via `ct list-changed --include-subcharts ...`.
This allows us to also test the subcharts. Do you agree ?

jlegrone:
I'm not convinced that the way things are laid out in source control in that repository is something we want to encode in chart-testing.

It seems like you'd be just as well served making the subcharts of "meta" charts subdirectories in the templates directory of the meta chart rather than dependencies (if they truly are not independently installable) or separate charts pulled in via helm's existing dependency management system. Note that [dependencies may be declared on local file paths within the same repo](https://github.com/helm/helm/blob/66d71811e843dce325feeda0a98d528fd425f92f/docs/helm/helm_dependency.md), eg:

```yaml
# requirements.yaml
dependencies:
- name: nginx
  version: "1.2.3"
  repository: "file://../dependency_chart/nginx"
```

Another workaround, if you want to keep the same directory and chart structure, may be to specify each `<meta_chart>/charts` path as a chart-dir passed to `ct`.

unguiculus:
I agree with @jlegrone and consider this non-standard. I'd suggest using dependencies via `requirements.yaml` which would also allow you to evolve your charts independently. Plus, it makes things explicit for the user.

auhlig:
Thanks for your input @unguiculus @jlegrone.
Umbrella Helm charts that can contain multiple subchart is a standard pattern as per [offical documentation](https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#complex-charts-with-many-dependencies). It explicitly mentions our [helm-charts](https://github.com/sapcc/helm-charts) as a reference.  
Thus I think `--include-subcharts` introduces a valid fix when using umbrella charts.

jlegrone:
@auhlig can you confirm whether this works?

> Another workaround, if you want to keep the same directory and chart structure, may be to specify each `<meta_chart>/charts` path as a chart-dir passed to `ct`.

auhlig:
> @auhlig can you confirm whether this works?
> 
> > Another workaround, if you want to keep the same directory and chart structure, may be to specify each `<meta_chart>/charts` path as a chart-dir passed to `ct`.

That does not work unfortunately @jlegrone  :-(

PR #164, opened 2019-07-17
Only print configuration if --debug flag is present.

tompowell-dd:
The current behavior prints HelmRepoExtraArgs to stdout. This can possibly include sensitive details (including helm repo logins). This PR hides configuration info behind the --debug flag. 

**Which issue this PR fixes**: None.


jlegrone:
I talked with @tompowell-dd offline, and I think a workaround is to just configure repositories before running `ct`. Repo configuration is really just part of `ct` as a convenience and could feasibly be removed in v3.

PR #167, opened 2019-07-24
Always use strict helm linting.

### LINKED ISSUE #166: Possible switch/config option to allow helm lint with "--strict"
There should be a switch or a config option that allows for the lint to use `helm lint --strict` to ensure a failure if there are missing values.

https://github.com/helm/chart-testing/blob/master/pkg/tool/helm.go#L53

Perhaps even just changing the line to
```golang
return h.exec.RunProcess("helm", "lint", "---strict", chart, values)
```

nicksunday:
**What this PR does / why we need it**: Strict helm linting does a more thorough validation of Chart configuration

**Which issue this PR fixes** : fixes #166

### DIFF
--- a/pkg/tool/helm.go
+++ b/pkg/tool/helm.go
@@ -50,7 +50,7 @@ func (h Helm) LintWithValues(chart string, valuesFile string) error {
 		values = []string{"--values", valuesFile}
 	}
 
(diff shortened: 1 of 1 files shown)


nicksunday:
Thanks for the feedback. I'll close this then.

PR #180, opened 2019-10-24
Prepare release v2.4.0

### LINKED ISSUE #179: Proposal: update docker image on quay.io
We have helm 2.14.1 in Dockerfile and 2.13.1 in the latest docker image in quay.io.

Helm 2.13.1 has some bugs that are affecting the github.com/helm/charts tests. For example, this issue: https://github.com/helm/charts/issues/9241 is affecting builds.

Could you please update the image to use at least the latest 2.14.x version of Helm?

EricLemieux:
Updates versions in example docs in order to prepare a new release that would contain changes since `v2.3.3`

**Which issue this PR fixes**
There is no formal issue number, but it would resolve an issue reported in [slack](https://kubernetes.slack.com/archives/CHK6WB5R8/p1571917162009500), 
```
vas 07:39
Hey, is there a plan to release a new chart-testing image with a newer version of Helm that contains this fix: https://github.com/helm/helm/pull/5112?
The latest image is 3.3.2, which has Helm 2.13.1, and it appears that this does not have the fix in it.
```
EDIT: Fixes #179 

Apologies if this is jumping the gun on a release, I am not too familiar with your standard process.

### DIFF
--- a/examples/docker-for-mac/my_test.sh
+++ b/examples/docker-for-mac/my_test.sh
@@ -18,7 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly IMAGE_TAG=v2.3.3
+readonly IMAGE_TAG=v2.4.0
 readonly IMAGE_REPOSITORY="quay.io/helmpack/chart-testing"
 
 main() {
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Run the script specifying the version for the new release.
 ./tag.sh --tag <release_version>
 ```
 
-Versions must start with a lower-case `v`, e. g. `v2.3.3`.
+Versions must start with a lower-case `v`, e. g. `v2.4.0`.
 
 
 ## Supported versions
--- a/examples/gke/Dockerfile
+++ b/examples/gke/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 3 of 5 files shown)


EricLemieux:
This is a full comparison of everything different between `v2.3.3` and `v2.4.0`
https://github.com/helm/chart-testing/compare/v2.3.3...EricLemieux:prepare-v2.4.0

rimusz:
LGTM

examples/docker-for-mac/my_test.sh
```diff
@@ -18,7 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly IMAGE_TAG=v2.3.3
+readonly IMAGE_TAG=v2.4.0
```
cpanato: this version is not available yet

EricLemieux: Reading this doc I was under the impression that the docs should be updated before the creation of the tag
https://github.com/helm/chart-testing#releasing

Should the tag be created first? Can I even create a tag?

unguiculus: It's in preparation for this version. So, I this is fine.

unguiculus: Tag creation is part of the release process. You can't do that.

scottrigby:
Thank you, merging now as part of the release window 🍺

PR #185, opened 2019-11-07
Bump helm to v2.16.0

EricLemieux:
**What this PR does / why we need it**:
Bumps the helm version to `v2.16.0`

**Special notes for your reviewer**:
I saw that the `v3.0.0` pr is in progress, not sure if you were planning on continuing to support 2.x.x after that is merged.

PR #187, opened 2019-11-12
Bump helm to v2.16.1

EricLemieux:
Bumps the helm version to `v2.16.1`

This is a followup to #185, where we decided that `v2.16.0` wasn't stable enough and opted to wait until `v2.16.1` was released.


EricLemieux:
@scottrigby could I get you to review this as a follow up to #185

PR #190, opened 2019-12-04
List changed fixes

jeff-knurek:
Using `list-changed` caused a lot of confusion when trying to get the structure/config correct. This PR is to help make it more understood.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -705,7 +705,7 @@ func (t *Testing) ComputeChangedChartDirectories() ([]string, error) {
 				changedChartDirs = append(changedChartDirs, chartDir)
 			}
 		} else {
-			fmt.Printf("Directory '%s' is no chart directory. Skipping...", chartDir)
+			fmt.Printf("Directory '%s' is not a valid chart directory. Skipping...\n", dir)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/chart/chart.go
```diff
@@ -705,7 +705,7 @@ func (t *Testing) ComputeChangedChartDirectories() ([]string, error) {
 				changedChartDirs = append(changedChartDirs, chartDir)
 			}
 		} else {
-			fmt.Printf("Directory '%s' is no chart directory. Skipping...", chartDir)
+			fmt.Printf("Directory '%s' is not a valid chart directory. Skipping...\n", dir)
```
jeff-knurek: When the tool found non-chart files changed, it output as:
```
Directory '' is no chart directory. Skipping...Directory '' is no chart directory. Skipping...Directory '' is no chart directory. Skipping...Directory '' is no chart directory. Skipping...
```
This was extremely confusing.

pkg/util/util.go
```diff
@@ -131,6 +131,7 @@ func (u ChartUtils) LookupChartDir(chartDirs []string, dir string) (string, erro
 		for {
 			chartYaml := path.Join(currentDir, "Chart.yaml")
 			parent := path.Dir(path.Dir(chartYaml))
+			chartDir = path.Join(chartDir) // removes any trailing slash from the dir
```
jeff-knurek: if you try setting `chart-dirs` as `./`, the tool gets stuck in an infinite loop. _(even more confusing because `.` works as expected)_ 
This was because of later in the code, 
`if (joined == chartDir) || strings.HasPrefix(relativeDir, "..") {` 
would check if `.` == `./`. 

I didn't test, but i'm pretty sure it was a similar problem with the `(parent == chartDir)` check as well.

unguiculus:
Thanks 🎉

PR #191, opened 2019-12-08
use single RUN in dockerfile to reduce layers & updated components

monotek:
Signed-off-by: André Bauer <monotek23@gmail.com>

* use single RUN in dockerfile to reduce layers
* use shell script docker-install.sh in RUN command to keep steps more readable
* updated helm, yamale & yamllint
* send readlink error messag to /dev/null in build.sh
* use curl -f to catch 404 errors (and -sS to be silent but show errors)


scottrigby:
@monotek 👋  just circling back to this now. I'm not sure why we'd want to do most of what's in this old PR. Did you want to close it?

Generally we want separate layers from least likely to more likely to change, to take advantage of caching the preceding layers that change less frequently

The install script is also atypical, except for runtime operations (ENTRYPOINT).

But I can certainly get behind `curl -sSfL` 👍

monotek:
Hi :)

If you don't like to merge it it's ok for me to close it. Motivation was to minimize number of layers.

I prefer install scripts to get around lots of " && /" to separate commands in single RUN instruction. Makes the install file more readable to me as the dockerfile, but yes, that's also bit opinionated.

PR #184, opened 2019-11-03
Add support for Helm 3

unguiculus:

PR #203, opened 2020-02-07
allow to skip namespace creation

torstenwalter:

In some environments it's not reasonable to allow creation and deletion of namespaces. This flag allows to use chart testing with already existing namespaces.

It's missing tests. Just want to get feedback if this would be a useful addition.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -541,8 +541,10 @@ func (t *Testing) doInstall(chart *Chart) error {
 			namespace, release, releaseSelector, cleanup := t.generateInstallConfig(chart)
 			defer cleanup()
 
-			if err := t.kubectl.CreateNamespace(namespace); err != nil {
-				return err
+			if t.config.SkipNamespaceCreation == false {
+				if err := t.kubectl.CreateNamespace(namespace); err != nil {
+					return err
+				}
 			}
 			if err := t.helm.InstallWithValues(chart.Path(), valuesFile, namespace, release); err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


pkg/chart/chart.go
```diff
@@ -541,8 +541,10 @@ func (t *Testing) doInstall(chart *Chart) error {
 			namespace, release, releaseSelector, cleanup := t.generateInstallConfig(chart)
 			defer cleanup()
 
-			if err := t.kubectl.CreateNamespace(namespace); err != nil {
```
jlegrone: I wonder if adding a check to see if the namespace already exists before creation would handle your use case without adding a new flag?

torstenwalter: That would be an option, too. In that case it should not try to delete the namepace at the end.

jlegrone: Actually if I remember correctly "do not create or delete namespace" was exactly the intended behavior of https://github.com/helm/chart-testing/pull/59. If passing the `namespace` flag can currently cause an existing namespace to be deleted I view that as a bug to be patched.

jlegrone: Yes looks like this was changed recently. https://github.com/helm/chart-testing/pull/184/files#diff-326b04dcc669c6ac56b6ea231f240fb2R584-R586

cc @unguiculus

unguiculus: Correct. This should already be supported. I'll have to look into this.

torstenwalter: @unguiculus Thank you for looking into it. In the source code it looks like as if this feature has been removed.

unguiculus: A regression was introduced due to Helm 3 no longer creating namespaces. This caused namespaces to always be created. #205 fixes this.

torstenwalter: Thanks fo looking into it. I am closing this in favor of your PR.

PR #205, opened 2020-02-14
Don't create provided namespace

unguiculus:
Adding support for Helm 3 introduced creating namespaces
because Helm does not do this automatically anymore.
However, a regression was introduced that always creates
namespaces, even if a dedicated namespace is provided via
CLI flag.

Signed-off-by: Reinhard Naegele <unguiculus@gmail.com>


unguiculus:
@jlegrone Whenever we create a namespace, we should also delete it again. That's what's the case here. So, this is correct IMO.

PR #207, opened 2020-02-19
Update Helm to v3.1.1

prehor:
Helm v3.0.1 lint on my library chart produces nested file error which is fixed in v3.1.0.


unguiculus:
Thanks. I just bumped it again to v3.1.1.

rimusz:
LGTM

PR #229, opened 2020-04-21
Add since flag for changing the reference branch for git merge-base

### LINKED ISSUE #64: Add option to git diff between last 2 local commits

marc-sensenich:
This adds a `since` flag that allows for the changing of the git reference for `git merge-base`. This is desired so that `ct list-changed` will allow for referencing the previous commit after a merge to master for CI/CD; for example

```sh
 ct list-changed --config .ct/config.yaml 
>>> git rev-parse --is-inside-work-tree
>>> git merge-base origin/master HEAD~1
>>> git diff --find-renames --name-only <commit> -- stable
```

**Which issue this PR fixes** 
Closes #64 

Closes #159 by picking up the work


monotek:
Thanks 😎

PR #233, opened 2020-06-17
Update Helm to 2.16.9

cpanato:
- update HELM_VERSION to v2.16.9
- alpine to 3.12
- kubectl to 1.18.6
- YAMALE to3.0.2
- YAML_LINT to 1.24.2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.11
+FROM alpine:3.12
 
 RUN apk --no-cache add \
     curl \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,31 +1,31 @@
-FROM alpine:3.11
+FROM alpine:3.12
 
 RUN apk --no-cache add \
     curl \
     git \
     libc6-compat \
     openssh-client \
-    python \
+    python3 \
```
unguiculus, line 8: Just to make sure. Did you test this? Are the tools compatible? How about the other Python dependencies: Are those still fine?

smlx:
Hi, just wanted to check on the status of this PR.

I'm hitting problems with the old version of helm in the chart-testing image that the chart-testing action uses by default. Specifically, the `lookup` function is failing validation (https://github.com/helm/helm/issues/7955). This is fixed in helm 3.2.x.

Dockerfile
cpanato: sorry for the delay, I've run the commands and after my latest change seems to work

cpanato:
@unguiculus @davidkarlsen followed the approach that we have for master branch

PR #239, opened 2020-07-16
appVersion should be optional

billimek:

This change makes the `appVersion` field in `Chart.yaml` an optional field, to match the [helm
documentation](https://helm.sh/docs/topics/charts/#the-appversion-field):
image


billimek:
@scottrigby or @unguiculus would you please take a look whenever convenient?

PR #247, opened 2020-07-28
Migrate chart-testing formula to homebrew/core

crunchtime-ali:
Signed-off-by: Alexander Zigelski <ali@crunchtime.dev>

Congratulations! 🎉 The formula `chart-testing` has been migrated to `homebrew-core` 🍺 
This means that the `homebrew` community will co-author the formula and the formula is even easier to install (via `brew install chart-testing`).

**What this PR does**:

This PR changes the instructions on how to install `chart-testing` on macOS. It also removes the tap configuration from `goreleaser` which prevents your own tap to be updated. The formula will be updated by the Homebrew community (e.g. automatically via `brew bump-formula-pr chart-testing $VERSION` which is triggered automatically by GitLab CI action in one of my private repos).
If you plan to automate `chart-testing` GitHub releases we can also include a GitHub action that automatically create a PR against `homebrew-core`. You can see an example on how to accomplish this in the `release.yaml` here https://github.com/argoproj/argo-cd/pull/3979/files

Here is the accompanying PR in your tap repo: https://github.com/helm/homebrew-tap/pull/1

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -56,18 +56,3 @@ dockers:
     extra_files:
       - etc/chart_schema.yaml
       - etc/lintconf.yaml
-brews:
-  - github:
-      owner: helm
-      name: homebrew-tap
-    commit_author:
-      name: helm-bot
-      email: helm-bot@users.noreply.github.com
-    folder: Formula
-    homepage: https://github.com/helm/chart-testing/
-    description: Testing and linting Helm charts
-    install: |
-      bin.install "ct"
-      etc.install "etc" => "ct"
-    test: |
-      system "#{bin}/ct version"
--- a/README.md
+++ b/README.md
@@ -37,7 +37,6 @@ available tags [here](https://quay.io/repository/helmpack/chart-testing?tab=tags
 ### Homebrew
 
 ```console
-$ brew tap helm/tap
 $ brew install chart-testing
 ```
 


crunchtime-ali:
I don't think my change caused the build to fail. Can someone kick the CI server to retry? Maybe @scottrigby ?

unguiculus:
I don't see any advantage here. I'd rather keep things under our control. Why would it make sense to rely on a CI job in one of your private repos?

crunchtime-ali:
> I don't see any advantage here. I'd rather keep things under our control. Why would it make sense to rely on a CI job in one of your private repos?

the advantages are:
- the binaries are bottled for the current macOS version and the former two
- the community makes sure that the formula and `chart-testing` itself works with the new architecture (welcome Apple Silicone) without you lifting a finger (see https://github.com/Homebrew/brew/issues/7857)
- you can check out installation analytics for the formula if you are interested in it

Of course, you don't want to rely on a CI job in one of my public repos but you can rely on the active Homebrew community to keep your formula in the best shape possible while you worry about the `chart-testing` project itself. 
Since `chart-testing` releases are not automated there was no point in adding a CI job to this repo. If you are interested in switching the build and release process to GitHub actions I could prepare another PR independent of this one.

If you still want to keep everything under your own belt feel free to close this PR.

unguiculus:
@crunchtime-ali Thanks for clarifying.

unguiculus (approved):

claui:
> * the community makes sure that the formula and `chart-testing` itself works with the new architecture (welcome Apple Silicone)

We only track the top 200 formulae for now, which doesn’t include chart-testing. However, we do track two of your dependencies, `go` and `helm`.

As soon as you see both `go` and `helm` listed 🥉 (or better) on https://github.com/Homebrew/brew/issues/7857, chances are `chart-testing` is going to work just fine, too.

Issue #251, opened 2020-08-05
type field missing from chart schema


smlx:
**Is this a request for help?**: No

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

**Version of Helm and Kubernetes**:

```
$ helm version
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"clean", GoVersion:"go1.13.12"}

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.1", GitCommit:"7879fc12a63337efff607952a323df90cdc7a335", GitTreeState:"clean", BuildDate:"2020-04-08T17:38:50Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

$ ct version
Version:	 v3.0.0
Git commit:	 50db473a1e68c605b18d82f019d83ea401542213
Date:		 2020-07-24T20:35:05Z
License:	 Apache 2.0
```

**What happened**:
I created a PR which ran the chart-testing github action (https://github.com/helm/chart-testing-action). My chart failed with:
```
Error validating data '/workdir/charts/lagoon-core/Chart.yaml' with schema '/etc/ct/chart_schema.yaml'
	type: Unexpected element
```

**What you expected to happen**:
I expected my chart to pass the `ct lint` check.

**How to reproduce it** (as minimally and precisely as possible):
Run the chart-testing github action against a chart which has a `type` field.

**Anything else we need to know**:
n/a

estahn:
The following worked for me on MacOS:
```bash
rm -rf /usr/local/etc/ct
brew reinstall chart-testing
```

PR #250, opened 2020-08-05
Bump docker image helm verison to 3.2.4

smlx:
**What this PR does / why we need it**: Bumps the helm version in the docker image to the latest 3.2.4

**Which issue this PR fixes** : fixes #249 

**Special notes for your reviewer**: This tool is great, thanks!

PR #234, opened 2020-06-17
update helm to 3.2.4 / alpine to 3.12

cpanato:
**What this PR does / why we need it**:
- update HELM_VERSION to v3.2.4
- alpine to 3.12

in the dockerfile:
- update python to use version 3

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

PR #253, opened 2020-08-05
Bump docker image versions

smlx:
**What this PR does / why we need it**: This PR bumps the version of the base image, and the tools within it, to their latest versions.

**Which issue this PR fixes**: supersedes #234 

**Special notes for your reviewer**: I notice that #234 has a request for extra testing, so I've built an image that contains #250, #251, and this PR. It's available [here](https://hub.docker.com/layers/smlx/chart-testing/v3.0.0-patch.1/images/sha256-b8f60684a36bd422a1fe9a8e97cf336584118f89f409a4b99114a2945cb1e5c1?context=explore). I've also tested this image [here](https://github.com/amazeeio/charts/runs/949550295) in the chart-testing lint action.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,22 @@
-FROM alpine:3.11
+FROM alpine:3.12
 
 RUN apk --no-cache add \
     curl \
     git \
     libc6-compat \
     openssh-client \
-    python \
-    py-crcmod \
-    py-pip && \
-    pip install --upgrade pip==18.1
+    py3-pip \
+    py3-wheel \
+    python3 \
+    && pip install --upgrade pip==20.2.1
 
 # Install a YAML Linter
-ARG yamllint_version=1.21.0
+ARG yamllint_version=1.24.2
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,22 +1,22 @@
...
     curl \
     git \
     libc6-compat \
     openssh-client \
-    python \
-    py-crcmod \
-    py-pip && \
-    pip install --upgrade pip==18.1
+    py3-pip \
+    py3-wheel \
```
unguiculus, line 9: I'm not a Python expert. Why would we want `py3-wheel`? The image builds fine without it.

smlx: The reason to install `wheel` is to avoid this warning that you'll see if you run `pip` and don't have `wheel` installed:
```
Using legacy 'setup.py install' for pyyaml, since package 'wheel' is not installed.
```
For an overview of what wheels are and why they're better than `setup.py`, there's a reasonable overview [here](https://realpython.com/python-wheels/).

PR #260, opened 2020-08-14
 Make home URL optional and support dependencies

minhdanh:
`home` is optional according to Helm's docs: https://helm.sh/docs/topics/charts/
```
home: The URL of this projects home page (optional)
```

Chart.yaml also has optional declaration for dependencies:
```
dependencies: # A list of the chart requirements (optional)
  - name: The name of the chart (nginx)
    version: The version of the chart ("1.2.3")
    repository: The repository URL ("https://example.com/charts") or alias ("@repo-name")
    condition: (optional) A yaml path that resolves to a boolean, used for enabling/disabling charts (e.g. subchart1.enabled )
    tags: # (optional)
      - Tags can be used to group charts for enabling/disabling together
    enabled: (optional) Enabled bool determines if chart should be loaded
    import-values: # (optional)
      - ImportValues holds the mapping of source values to parent key to be imported. Each item can be a string or pair of child/parent sublist items.
    alias: (optional) Alias to be used for the chart. Useful when you have to add the same chart multiple times
```

### DIFF
--- a/etc/chart_schema.yaml
+++ b/etc/chart_schema.yaml
@@ -7,6 +7,7 @@ description: str()
 keywords: list(str(), required=False)
 sources: list(str(), required=False)
 maintainers: list(include('maintainer'), required=False)
+dependencies: list(include('dependency'), required=False)
 icon: str(required=False)
 engine: str(required=False)
 condition: str(required=False)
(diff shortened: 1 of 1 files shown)


etc/chart_schema.yaml
```diff
@@ -20,3 +21,17 @@ maintainer:
...
   email: str(required=False)
   url: str(required=False)
+---
+dependency:
+  name: str()
+  version: str()
+  repository: str()
+  condition: str(required=False)
+  tags: list(str(), required=False)
+  enabled: bool()
```
smlx, line 31: This field is optional according to https://helm.sh/docs/topics/charts/#the-chartyaml-file

minhdanh: @smlx Thanks. I've just updated that

minhdanh:
@smlx Can we have this merged this soon?

smlx (approved):
I'd love to, but I don't have a commit bit. That said, LGTM.

unguiculus (approved):

PR #269, opened 2020-09-09
Add flag for disabling config printing

unguiculus:
By default, ct prints the config on startup. This can be problematic
because it may contain senstivie data when helm-repo-extra-args contains
passwords. The new flag enables turning off config printing.

### DIFF
--- a/ct/cmd/root.go
+++ b/ct/cmd/root.go
@@ -92,7 +92,9 @@ func addCommonLintAndInstallFlags(flags *pflag.FlagSet) {
 		(e.g. 'myrepo=--username test --password secret'). May be specified
 		multiple times or separate values with commas`))
 	flags.Bool("debug", false, heredoc.Doc(`
-		Print CLI calls of external tools to stdout (Note: depending on helm-extra-args
-		passed, this may reveal sensitive data)`))
-	flags.Bool("print-config", true, "Prints the configuration to stdout")
+		Print CLI calls of external tools to stdout (caution: setting this may
+		expose sensitive data when helm-repo-extra-args contains passwords)`))
+	flags.Bool("print-config", false, heredoc.Doc(`
+		Prints the configuration to stdout (caution: setting this may
+		expose sensitive data when helm-repo-extra-args contains passwords)`))
 }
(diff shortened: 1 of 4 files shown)


ct/cmd/root.go
```diff
@@ -94,4 +94,5 @@ func addCommonLintAndInstallFlags(flags *pflag.FlagSet) {
 	flags.Bool("debug", false, heredoc.Doc(`
 		Print CLI calls of external tools to stdout (Note: depending on helm-extra-args
 		passed, this may reveal sensitive data)`))
+	flags.Bool("print-config", true, "Prints the configuration to stderr")
```
mattfarina, line 97: This flag seems a bit not intuitive to me. The default is to print the config (this has a true value). So, you don't pass in the flag to skip it. You have to say it's equal to false. Would it be better to make this `--skip-print-config` with a value of false? Then use the `not` of this value for the `LoadConfiguration` function?

unguiculus: Good point. not sure what's better. Ideally, I'd like to change the default to false. I was hesitant to do so. Would that be considered a breaking change?

mattfarina: Changing a default value is considered a breaking change. The helm client would only do that on a major version change.

scottrigby: This is fun, because we already have a `printConfig` arg in `LoadConfiguration()`, but had no actual config/flag for it 😄 

Given the security implications I think we should change chart-testing to print not the config values by default, and whatever flag we decide is most intuitive to expose to end uses add a note in the flag help/description what you said initially @unguiculus:

> caution: setting this may expose sensitive data when helm-repo-extra-args contains passwords

I'm a bit torn on the semver interpretation here. If it's considered a security bug, then changing output wouldn't require MAJOR. In that case, the additional flag to output it if desired would only require MINOR. Thoughts on this?

unguiculus: I agree with @scottrigby. After all it's only a change in log output, not in functionality.

scottrigby: Cool. Whatever we decide with semver, not printing this by default is the right thing to do

scottrigby (approved):

Issue #276, opened 2020-09-25
[BUG REPORT] ct list-changed  returns additional config as  output


chukka:
**Is this a request for help?**: **No**

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):**BUG**

```
ct version - 3.1.0

**What happened**:
Upgraded to CT - 3.1.0  on Mac and found that `ct list-changed` gives `Using config file: test/ct.yaml` in the output as first result
```
$ ct list-changed --config test/ct.yaml
Using config file: test/ct.yaml
stable/artifactory-ha
stable/artifactory

**What you expected to happen**: Expected output
$ ct list-changed --config test/ct.yaml
stable/artifactory-ha
stable/artifactory

**How to reproduce it** (as minimally and precisely as possible):
Use CT- 3.1.0 , change a chart locally and run `ct list-changed --config test/ct.yaml` using a config file it gives additional result of 
Using config file: test/ct.yaml
**Anything else we need to know**:

chukka:
@unguiculus @davidkarlsen  Looks like a regression issue ?

unguiculus:
Indeed, that's a regression. Sorry about that. Fix is on its way.

PR #277, opened 2020-09-25
Print config to stderr

unguiculus:
Fixes a regression when the config file used is printed
to stdout even when the list-changed command is run. It is
now only printed when --print-config=true. Also, config printing
is now done to stderr so the flag can also be used in combination
with the list-changed command.

Fixes: #276


davidkarlsen (approved):

pkg/util/util.go
```diff
@@ -210,12 +211,12 @@ func BreakingChangeAllowed(left string, right string) (bool, error) {
 	return !minor, err
 }
 
-func PrintDelimiterLine(delimiterChar string) {
+func PrintDelimiterLine(w io.Writer, delimiterChar string) {
```
mattfarina, line 214: This signature changing breaks the API from a semver perspective. If this were the Helm client we would have kept the existing function, deprecated it, created a new function doing what we want, and then used that all over.

unguiculus: chart-testing was never meant to be used as API and it is not advertized as such. Admittedly, we don't hide the code in an internal package. I'm gonna hold back merging this until we get consensus on how to proceed.

unguiculus: OK, so I went ahead and added a new function so we can get this out soon. We should still have the discussion on how we want to generally handle this.

pkg/util/util.go
```diff
@@ -211,11 +212,15 @@ func BreakingChangeAllowed(left string, right string) (bool, error) {
 }
 
 func PrintDelimiterLine(delimiterChar string) {
```
scottrigby, line 214: We may as well note this func is deprecated
```suggestion
// Deprecated, to be removed in v4
func PrintDelimiterLine(delimiterChar string) {
```

pkg/util/util.go
scottrigby: I think this makes sense 👍  We don't explicitly advertise exported semver compliance for chart-testing, but I think it may be implied 🤔  For now the new function seems like a safe approach to fix quickly without making a breaking decision. I'm making a review note about adding a deprecation comment.

scottrigby (approved):
lgtm 👍

pkg/util/util.go
unguiculus: I had thought about it and didn't do it. But I guess it makes sense.

scottrigby: yep probably fine either way (I "approved" with this optional suggestion). Accept if you want to, but either way 🚀 🙂

unguiculus: Fixed

PR #283, opened 2020-10-05
Support execution of additional commands during ct lint

torstenwalter:

This is an alternative design to #279 which not only support `helm unittest`, but running any command as part of `ct lint`.
@unguiculus mentioned in slack that you would prefer that instead of pulling in additional dependencies.

I think that's totally valid and also solve many issue. When using #279 I noticed for example that it would be nice if I could pass additional parameters to the command. In that case I needed to add `--helm3`.

Given that helm unittest is installed locally or mounted into the chart testing container one could use this to run helm unittest for each chart with this configured in `ct.yaml`:

```
additional-commands:
 - helm unittest --helm3 -f tests/*.yaml {{ .Path }}
```

The command is treated as a go template. Like this it's possible to get the path to the chart as in the example. It would also be open for further extension if needed.

One idea would be to pass the information if it's an helm 2 or helm 3 chart.

This works fine when ct is run locally and the toll is installed on your machine. If you want it in the container then the tool needs to be present there. There are multiple option to achieve this:

- built a custom image, which contains the tool
- mount a directory in case the tool has no dependencies
  PR helm/chart-testing-action/pull/54 adds support for it in chart-testing-action.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM alpine:3.12
 
 RUN apk --no-cache add \
+    bash \
     curl \
     git \
     libc6-compat \
--- a/ct/cmd/lint.go
+++ b/ct/cmd/lint.go
@@ -70,7 +70,10 @@ func addLintFlags(flags *flag.FlagSet) {
 	flags.Bool("validate-yaml", true, heredoc.Doc(`
 			Enable linting of 'Chart.yaml' and values files (default: true)`))
 	flags.StringSlice("additional-commands", []string{}, heredoc.Doc(`
-			Additional command to run per chart (default: [])`))
+            Additional commands to run per chart (default: [])
+            Commands will be executed in the same order as provided in the list and will
+            be rendered with go template before being executed.
+            Example: "helm unittest --helm3 -f tests/*.yaml {{ .Path }}"`))
 }
 
 func lint(cmd *cobra.Command, args []string) error {
--- a/ct/cmd/lint.go
+++ b/ct/cmd/lint.go
@@ -69,6 +69,8 @@ func addLintFlags(flags *flag.FlagSet) {
 			Enable schema validation of 'Chart.yaml' using Yamale (default: true)`))
 	flags.Bool("validate-yaml", true, heredoc.Doc(`
 			Enable linting of 'Chart.yaml' and values files (default: true)`))
+	flags.StringSlice("additional-commands", []string{}, heredoc.Doc(`
+			Additional command to run per chart (default: [])`))
 }
 
 func lint(cmd *cobra.Command, args []string) error {
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -261,7 +261,7 @@ func NewTesting(config config.Configuration) (Testing, error) {
 		git:              tool.NewGit(procExec),
 		kubectl:          tool.NewKubectl(procExec),
 		linter:           tool.NewLinter(procExec),
-		cmdExecutor:      tool.NewTester(procExec),
+		cmdExecutor:      tool.NewCmdTemplateExecutor(procExec),
 		accountValidator: tool.AccountValidator{},
 		directoryLister:  util.DirectoryLister{},
 		chartUtils:       util.ChartUtils{},
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -131,6 +131,13 @@ type Linter interface {
 	Yamale(yamlFile string, schemaFile string) error
 }
 
+// CmdExecutor is the interface
+//
+// RunCommand renders cmdTemplate as go template using data and executes the resulting command
+type CmdExecutor interface {
+	RunCommand(cmdTemplate string, data interface{}) error
+}
+
 // DirectoryLister is the interface
 //
 // ListChildDirs lists direct child directories of parentDir given they pass the test function
@@ -224,6 +231,7 @@ type Testing struct {
 	kubectl                  Kubectl
 	git                      Git
 	linter                   Linter
+	cmdExecutor              CmdExecutor
 	accountValidator         AccountValidator
 	directoryLister          DirectoryLister
 	chartUtils               ChartUtils
@@ -253,6 +261,7 @@ func NewTesting(config config.Configuration) (Testing, error) {
 		git:              tool.NewGit(procExec),
 		kubectl:          tool.NewKubectl(procExec),
 		linter:           tool.NewLinter(procExec),
+		cmdExecutor:      tool.NewTester(procExec),
 		accountValidator: tool.AccountValidator{},
 		directoryLister:  util.DirectoryLister{},
 		chartUtils:       util.ChartUtils{},
@@ -446,6 +455,13 @@ func (t *Testing) LintChart(chart *Chart) TestResult {
 
 	if t.config.ValidateMaintainers {
 		if err := t.ValidateMaintainers(chart); err != nil {
+			result.Error = err
+			return result
+		}
+	}
+
+	for _, cmd := range t.config.AdditionalCommands {
+		if err := t.cmdExecutor.RunCommand(cmd, chart); err != nil {
 			result.Error = err
 			return result
 		}
--- a/pkg/tool/cmdexecutor.go
+++ b/pkg/tool/cmdexecutor.go
@@ -22,9 +22,5 @@ func (t CmdTemplateExecutor) RunCommand(cmdTemplate string, data interface{}) er
 	var b strings.Builder
 	err := template.Execute(&b, data)
 	renderedCommand := b.String()
-	split := strings.Split(renderedCommand, " ")
-	if err != nil {
-		return err
-	}
-	return t.exec.RunProcess(split[0], split[1:])
+	return t.exec.RunProcess("sh", "-c", renderedCommand)
 }
--- a/pkg/tool/cmdexecutor.go
+++ b/pkg/tool/cmdexecutor.go
@@ -4,15 +4,18 @@ import (
 	"strings"
 	"text/template"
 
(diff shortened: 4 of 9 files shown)


torstenwalter:
Here is an example usage:

- installing the plugin https://github.com/torstenwalter/jenkins-helm-charts/blob/unit-tests-improved/.github/workflows/lint-test.yaml#L19-L26
- mounting it into the container https://github.com/torstenwalter/jenkins-helm-charts/blob/unit-tests-improved/.github/workflows/lint-test.yaml#L35
- specifying the aditional command to run
https://github.com/torstenwalter/jenkins-helm-charts/blob/unit-tests-improved/ct.yaml#L7-L8

Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM alpine:3.12
 
 RUN apk --no-cache add \
+    bash \
```
torstenwalter, line 4: This is not really needed for this PR. I added it while I tested to install helm unittest in the container.
Looks like some plugins need bash to be able to use `helm plugin install`.

Adding it here makes it easier to extend the container.

Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM alpine:3.12
```
torstenwalter, line 1: Not sure if this will limit executables within the container or if debian slim would be a better option.

pkg/tool/cmdexecutor.go
```diff
@@ -0,0 +1,30 @@
...
+func (t CmdTemplateExecutor) RunCommand(cmdTemplate string, data interface{}) error {
+	var template = template.Must(template.New("command").Parse(cmdTemplate))
+	var b strings.Builder
+	err := template.Execute(&b, data)
+	renderedCommand := b.String()
+	split := strings.Split(renderedCommand, " ")
+	if err != nil {
+		return err
+	}
+	return t.exec.RunProcess(split[0], split[1:])
```
unguiculus, line 29: What if an argument or a path contains spaces? You should use a shell to run the process.

```suggestion
	return t.exec.RunProcess("sh", "-c", renderedCommand)
```

pkg/tool/cmdexecutor.go
```diff
@@ -0,0 +1,30 @@
...
+	"text/template"
+
+	"github.com/helm/chart-testing/v3/pkg/exec"
+)
+
+type CmdTemplateExecutor struct {
+	exec exec.ProcessExecutor
+}
+
+func NewTester(exec exec.ProcessExecutor) CmdTemplateExecutor {
```
unguiculus, line 14: Name not ideal

pkg/chart/chart.go
```diff
@@ -451,6 +460,13 @@ func (t *Testing) LintChart(chart *Chart) TestResult {
 		}
 	}
 
+	for _, cmd := range t.config.AdditionalCommands {
+		if err := t.cmdExecutor.RunCommand(cmd, chart); err != nil {
```
unguiculus, line 464: We may also want to pass in the config for templating

ct/cmd/lint.go
```diff
@@ -69,6 +69,8 @@ func addLintFlags(flags *flag.FlagSet) {
 			Enable schema validation of 'Chart.yaml' using Yamale (default: true)`))
 	flags.Bool("validate-yaml", true, heredoc.Doc(`
 			Enable linting of 'Chart.yaml' and values files (default: true)`))
+	flags.StringSlice("additional-commands", []string{}, heredoc.Doc(`
+			Additional command to run per chart (default: [])`))
```
unguiculus, line 73: We should add some more documentation for this.

unguiculus (changes requested):
cc @scottrigby @davidkarlsen

pkg/tool/cmdexecutor.go
torstenwalter: Do you have a suggestion?

torstenwalter: It's now `NewCmdTemplateExecutor`.

ct/cmd/lint.go
torstenwalter: Good point. I added some.

pkg/tool/cmdexecutor.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (t CmdTemplateExecutor) RunCommand(cmdTemplate string, data interface{}) error {
+	var template = template.Must(template.New("command").Parse(cmdTemplate))
+	var b strings.Builder
+	if err := template.Execute(&b, data); err != nil {
+		return err
+	}
+	renderedCommand := b.String()
+	return t.exec.RunProcess("sh", "-c", renderedCommand)
```
nvtkaszpir, line 27: This means given user needs to have a shell installed, which may be a problem on other platforms (especially on Windows).
I believe it should be a raw RunProcess, in that case someone can explicitly define how to run commands or other tool to invoke commands (for example in another container, but that would be container in container in some cases....).

torstenwalter: I would be fine with executing it without a shell @unguiculus what's your opinion?

torstenwalter: The windows argument only holds true if you execute chart testing locally. If you use the built container it's always linux and there using a shell could be preferable.

nvtkaszpir: One more thing, the 'sh' may run different shell than other people would expect - I refer bash vs dash in here, especially under os x.

torstenwalter: Most freedom would be to simply leave it out. I mean users could add `sh -c` in their additional command. I just want to get feedback from @unguiculus as he suggested to add it in the first place.

scottrigby:
Question: why build this into charts-testing?

As opposed to say just adding that line (or any other command for that matter) as a separate step in your CI?

I understand the `quintush/helm-unittest` plugin literally relates to "chart testing", but it's currently not related to this project at all apart from that. So this PR seems to be using the chart-testing command as a passthrough for literally any other command with no benefit I can see yet over calling those other commands separately. I could be wrong though, please let me know if there are other benefits.

If we did want to expand the scope of `helm/chart-testing` to support unit tests via that plugin, I would prefer expanding on your other PR https://github.com/helm/chart-testing/pull/279. These are just my initial thoughts. What do you think @unguiculus?

torstenwalter:
The benefit of putting it here is the logic to detect which charts have been modified. It allows to run commands for modified charts only.

Otherwise you would need to duplicate that in your CI pipeline.

Allowing commands over just supporting helm unittest allows users to extend even further. One idea would be to check for deprecated Kubernetes APIs.

pkg/tool/cmdexecutor.go
scottrigby: From https://github.com/helm/chart-testing/pull/283#discussion_r499725440

> What if an argument or a path contains spaces? You should use a shell to run the process.

These could then be wrapped in quotes, yes? Aren't we already properly wrapping quotes?

scottrigby:
That sounds like a good reason 👍

pkg/tool/cmdexecutor.go
unguiculus: You need to somehow separate command from args. If a user configures "sh -c 'mytool'" and you pass that as command, it wouldn't work.

Maybe have a look at https://github.com/mattn/go-shellwords, which would allow us to parse the commandline and run it using an OS-specific shell.

scottrigby: Yep 👍  [func Command(name string, arg ...string) *Cmd](https://golang.org/pkg/os/exec/#Command) needs a command `name` string, then variadic `args`.

Did a quick playground test with go-shellwords and seems to work as expected: https://play.golang.org/p/H9Q9MO_0e2R Nice find 😄

torstenwalter: @unguiculus Thanks for the hint. command and args are now spitted with   https://github.com/mattn/go-shellwords

torstenwalter:
Is there anything else, which should be changed?

pkg/tool/cmdexecutor.go
```diff
@@ -23,6 +24,12 @@ func (t CmdTemplateExecutor) RunCommand(cmdTemplate string, data interface{}) er
...
 	}
-	renderedCommand := b.String()
-	return t.exec.RunProcess("sh", "-c", renderedCommand)
+	rendered := b.String()
+
+	words, err := shellwords.Parse(rendered)
+	name, args := words[0], words[1:]
+	if err != nil {
+		return err
+	}
```
unguiculus, line 33:
```suggestion
	if err != nil {
		return err
	}
	name, args := words[0], words[1:]
```

unguiculus (changes requested):

pkg/tool/cmdexecutor.go
```diff
@@ -0,0 +1,35 @@
...
+	if err := template.Execute(&b, data); err != nil {
+		return err
+	}
+	rendered := b.String()
+
+	words, err := shellwords.Parse(rendered)
+	if err != nil {
+		return err
+	}
+	name, args := words[0], words[1:]
```
unguiculus, line 33: You should check the length of the slice. There might not be any args.

unguiculus (changes requested):
Could you add unit tests, please?

pkg/tool/cmdexecutor.go
torstenwalter: That's not an issue. `words[1:]` will be an empty slice in that case. I added a unit test for that:
https://github.com/torstenwalter/chart-testing/blob/additional-commands/pkg/tool/cmdexecutor_test.go#L29-L39

unguiculus (approved):

scottrigby (approved):
@torstenwalter this is great 💯 Thanks!

PR #290, opened 2020-10-28
Log invalid chart dirs to stderr

unguiculus:
If directories are deleted from a chart, ct detects them as invalid
chart directories. A log message is printed to stdout which messes up
the regular output. In order to fix this, the message is now logged to
stderr.

*Example:*

```
$ ct list-changed
Directory 'charts/mychart/templates/deleted' is not a valid chart directory. Skipping...
charts/mychart
```

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -17,6 +17,7 @@ package chart
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


scottrigby (approved):
💯

scottrigby:
In keeping with our last convo about output changes re stout/stderr ￼I would consider this a bug fix.

PR #294, opened 2020-11-03
Ignore stderr when parsing Helm version

unguiculus:
Helm v3.4.0 logs a warning about URL deprecation which causes semver
parsing to fail. We need to only read stdout and ignore stderr here.


chukka:
@unguiculus  are we planning for a new release with this fix ?

scottrigby (approved):
Good call 👍

PR #297, opened 2020-11-06
Move to gh-actions

### LINKED ISSUE #296: chore: Move from circle to gh-actions
Replace the current circle setup with gh-actions.
Have a build workflow, and a release-workflow which we can trigger by clicking.

davidkarlsen:
Fixes #296 

* Release is now a workflow dispatch
* Upgrade to helm v3.4.0

Note: For release to work the secrets for quay needs to be defined: QUAY_USERNAME + QUAY_PASSWORD
You can see the wf here: https://github.com/davidkarlsen/chart-testing/actions but I have a snag with the user initiated wf - it doesn't appear, which might be because I am on a fork or it's not defined in the default branch.

---
This change is [Reviewable](https://reviewable.io/reviews/helm/chart-testing/297)


davidkarlsen:
@unguiculus @scottrigby ptal?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Create GitHub Release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: false
+          prerelease: false
```
unguiculus, line 71: This should be removed. GoReleaser creates the release.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,71 @@
...
+          goimports -w -l .
+          go mod tidy
+          git diff --exit-code
+
+      - name: Login to registry
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
```
unguiculus, line 43: Previously, we had DOCKER_USERNAME and DOCKER_PASSWORD. I'd prefer these. We will have to add the secrets.

unguiculus (changes requested):
In general, I think it would be a good idea to have the same setup for both, chart-testing and chart-releaser. I went a different route for chart-releaser but don't really have strong opinions. Since chart-releaser's move to GitHub actions we've been having issues with Go modules because go.mod and go.sum would change during the build which I could not reproduce locally. This doesn't seem to be the case here. Have you experienced any such issues? If not, it may make more sense to proceed with your setup and adjust that of chart-releaser.

davidkarlsen:
> In general, I think it would be a good idea to have the same setup for both, chart-testing and chart-releaser. I went a different route for chart-releaser but don't really have strong opinions. Since chart-releaser's move to GitHub actions we've been having issues with Go modules because go.mod and go.sum would change during the build which I could not reproduce locally. This doesn't seem to be the case here. Have you experienced any such issues? If not, it may make more sense to proceed with your setup and adjust that of chart-releaser.

I haven't had problems with go.{mod,sum} changing unexpectedly (or at all) - so I though I'd leave them in as a security-measure - we can remove them if we see any trouble down the lane.

davidkarlsen:
@unguiculus wdyt - can we merge it now?

unguiculus (approved):

unguiculus:
I disabled the old Circle CI project.

PR #335, opened 2021-05-14
adding option to only show logs + events on failure (#134)

### LINKED ISSUE #134: Add option to hide pod logs/resource descriptions unless release tests fail
When testing even a single chart with multiple ci values files, the log output that `ct`produces is pretty large. When release tests fail, it would be much easier to figure out the cause if the log output from all the other successful releases could be filtered out.

deweya:
Closes #134.

Adds a new flag for install and lint-and-install called `--log-failed-only`. When applied, logs, events, and descriptions will only be logged when the installation/upgrade/tests fail. This will help prevent overly verbose output in the event users are trying to troubleshoot failures.

Signed-off-by: Austin Dewey <deweya964@gmail.com>

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -562,17 +562,25 @@ func (t *Testing) doInstall(chart *Chart) error {
 		// and be executed in reverse order after the loop.
 		fun := func() error {
 			namespace, release, releaseSelector, cleanup := t.generateInstallConfig(chart)
-			defer cleanup()
+			installFailed := false
+			defer cleanup(&installFailed)
 
 			if t.config.Namespace == "" {
 				if err := t.kubectl.CreateNamespace(namespace); err != nil {
+					installFailed = true
 					return err
 				}
 			}
 			if err := t.helm.InstallWithValues(chart.Path(), valuesFile, namespace, release); err != nil {
+				installFailed = true
 				return err
 			}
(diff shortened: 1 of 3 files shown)


pkg/chart/chart.go
```diff
@@ -650,19 +668,23 @@ func (t *Testing) testRelease(namespace, release, releaseSelector string) error
...
+			if !t.config.LogFailedOnly || (*failed && t.config.LogFailedOnly) {
+				t.PrintEventsPodDetailsAndLogs(namespace, releaseSelector)
+			}
 			t.helm.DeleteRelease(namespace, release)
 		}
 	} else {
 		release, namespace = chart.CreateInstallParams(t.config.BuildId)
-		cleanup = func() {
-			t.PrintEventsPodDetailsAndLogs(namespace, releaseSelector)
+		cleanup = func(failed *bool) {
```
unguiculus, line 684: Any reason why you use a pointer here?

deweya: Since the [defer](https://github.com/deweya/chart-testing/blob/feat/hide-logs/pkg/chart/chart.go#L566) takes place before we know if the test [failed or not](https://github.com/deweya/chart-testing/blob/feat/hide-logs/pkg/chart/chart.go#L573-L581), I thought it would have to be passed by reference

Issue #332, opened 2021-05-05
Log output for container in 'ct install' does not terminate


pathob:
**Is this a request for help?**: No

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Bug Report

**Version of Helm and Kubernetes**:

```
$ helm version
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"archive", BuildDate:"2020-07-28T00:00:00Z", GoVersion:"go1.15rc1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.17", GitCommit:"f00bd5fc5d1b9001e612847aa77f4e9c66758075", GitTreeState:"clean", BuildDate:"2021-03-22T23:03:44Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
$ ct version
Version:	 v3.3.1
Git commit:	 71d0e1e82c5c8b66ce4d9704426dc13b7075829d
Date:		 2020-11-30T20:10:04Z
License:	 Apache 2.0
```

**What happened**:

When running `ct install ...` the installation and tests do not terminate. The execution of ct gets stuck after:

```
<== Logs of init container crowd-uinddlbjpw-0
------------------------------------------------------------------------------------------------------------------------
>>> kubectl get pods crowd-uinddlbjpw-0 --no-headers --namespace crowd-uinddlbjpw --output jsonpath={.spec.containers[*].name}
------------------------------------------------------------------------------------------------------------------------
==> Logs of container crowd-uinddlbjpw-0
------------------------------------------------------------------------------------------------------------------------
>>> kubectl logs crowd-uinddlbjpw-0 --namespace crowd-uinddlbjpw --container crowd
```

The pod reaches running state and the displayed log is okay. But it just gets stuck there.

**What you expected to happen**:

As the installation (there are no actual tests yet) was successful, ct should just terminate successfully too.

[...]

PR #360, opened 2021-10-12
Set a request timeout on kubectl commands

smlx:

Simple kubectl commands such as get, describe, and logs should return
quickly, but kubectl does not set a timeout on requests by default.

This change sets a 30s timeout by default on such kubectl commands to
ensure that these don't hang forever if the kubernetes API stops
responding, which is quite possible in CI environments where jobs may
cancelled and kubernetes clusters torn down while ct is running.

This timeout value is configurable via the new `kubectl-timeout` option.

**Which issue this PR fixes**

Hopefully closes: #332

n/a


cpanato:
+1 to have that configurable

smlx:
Okay I added a configuration option for the kubectl timeout value.

davidkarlsen (approved):
lgtm

cpanato (approved):
thanks!

PR #388, opened 2022-02-01
Adding size labeler as a GH action

mattfarina:
The current automation for sizing PRs is run as a bot and
needs to be retired. This action with this configuration
will size PRs in the same manner the bot has.

**What this PR does / why we need it**: The PR sizing bot is going away and this PR provides a GitHub action that does the same thing.

**Special notes for your reviewer**: N/A


mattfarina:
Closing as we need another method to do labels. A github action doesn't appear to work due to permissions.

Issue #138, opened 2019-03-29
How to test charts in a local directory?


KIVagant:
**Is this a request for help?**:

Yes 
---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
No

**Version of Helm and Kubernetes**:

```
Client: &version.Version{SemVer:"v2.12.3", GitCommit:"eecf22f77df5f65c823aacd2dbd30ae6c65f186e", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.12.3", GitCommit:"eecf22f77df5f65c823aacd2dbd30ae6c65f186e", GitTreeState:"clean"}
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-04T04:48:55Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.8-eks-7c34c0", GitCommit:"7c34c0d2f2d0f11f397d55a46945193a0e22d8f3", GitTreeState:"clean", BuildDate:"2019-03-01T22:49:39Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
Version:	 unreleased
Git commit:	 unknown
Date:		 unknown
License:	 Apache 2.0
```

**What happened**:

```
$ ct lint-and-install   --chart-dirs .
Linting and installing charts...
Error loading configuration: 'chart_schema.yaml' neither specified nor found in default locations
```

**What you expected to happen**:

It should lint charts in the local directory. I have no idea how the `chart_schema.yaml` should look like.

**How to reproduce it** (as minimally and precisely as possible):

1. Create a dir.
2. Create a chart inside the dir.
3. Run `ct lint --chart-dirs .`

Thank you!

**Update** Same for `lintconf.yaml`. I believe the program should not require any specific configs to run it locally but just use some predefined settings.

jlegrone:
@KIVagant I agree that we should define defaults for both of these that are shipped with the binary.

In the mean time, you could either use the [docker image](https://github.com/helm/chart-testing#docker-image) (which also comes with `yamllint` and `yamale` installed) or use `ct` with the example config files found here:

[...]

PR #389, opened 2022-02-01
Update docs with example using ct with local repo

nvtkaszpir:

Just extend docs with common use case of how to test locally developed chart which is not under `charts/` subdirectory.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*:

Fixes: #138

Tested both commands locally. In theory it may need to the docstrings but I just thought README.md is good enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,24 @@ The following example show various way of configuring the same thing:
 
 #### CLI
 
+#### Remote repo
+
+With remote repo:
+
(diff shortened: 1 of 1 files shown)


cpanato (approved):
thanks for your PR

davidkarlsen (approved):

PR #396, opened 2022-02-18
introduce `helm-dependency-extra-args`

### LINKED ISSUE #368: Ability to add '--skip-refresh' to "helm dependency build"
**FEATURE REQUEST**

**What happened**:

When running chart-testing against multiple charts, the command `helm dependency build X` is run for each chart.  On each run, I see the following text
```
Hang tight while we grab the latest from your chart repositories...
```

We have many helm chart repositories and this refresh on each chart we test in our mono repo is the slowest part of our CI/CD since it involves external network calls.  Helm has an option, `--skip-refresh`, that can be passed to `dependency build`.  This will skip this refresh step and greatly speed up CI/CD in helm mono repo setups.

An option to tell chart-testing to run the helm refresh once, at the start of the command, then run with --skip-refresh on each sub chart.  This will make CI/CD fast by not running refresh on each chart, but only at the start.

Run chart-testing against multiple charts at once.

nazarewk:
TLDR; allows passing `--skip-refresh` to `helm dependency build`

Default behavior of `helm dependency build` is running equivalent of `helm repo update` before initiating packaging process. 
This makes sense for handling a single repository - you always get the latest version even if you forget about running `helm repo update` yourself.

Passing `--skip-refresh` turns off the step allowing the user to run `helm repo update` once for the whole pipeline. The default behavior becomes redundant when handling a second chart in the same CI pipeline and becomes *extremely* wasteful when running build on large number of charts in a single pipeline.

`ct lint` invokes that default behavior for every single repository it checks, making it unnecessarily wasteful for it's primary use case. Sane default would be recommending `ct` users to invoke `helm repo update` before running `ct lint`, but that would be a breaking change.

This PR allows us to still invoke the proper behavior while preserving backwards compatiblity for `ct` users.

**Which issue this PR fixes**: fixes #368

### Mocked test
I am not very proficient in mocking or Golang alltogether, but I believe `TestLintYamlValidation` works only by chance and my new `TestLintDependencyExtraArgs` contains proper invocation of `Mock.On()`.

Basically I could not get the `fakeMockLinter.AssertNumberOfCalls(t, "BuildDependenciesWithArgs", 1)` to count occurences at all.

Submitted test is what I have determined to work empirically. When I ran the test twice with just a single `.On()` invocation it threw an error, invoking `.On()` followed by `Repeatability = 1` allows a single call to the mocked method per test.

### Introducing new method
I am not sure what is the project's external compatibility policy. I've added a fresh method to preserve backwards compatiblity when using `ct` as a library, since the struct is public.


cpanato:
@nazarewk  thanks for this PR, can you check the ci failures?

nazarewk:
@cpanato done, looks like i was using `mock.Mock` instead of `*mock.Mock`.

nazarewk:
This time it should pass (i've run `go vet ./... && go test ./...` locally), looks like it was not the right change before.

cpanato (changes requested):
@nazarewk  thanks for the PR
can you generate the docs using the doc-gen? will be good as well to fetch the latest code from main

```
go build -o ct-bin ./ct/main.go

$ ./ct-bin doc-gen
Generating docs...
Done.
```

then commit the changes?
thanks!

nazarewk:
@cpanato docs are generated

cpanato (approved):

cpanato:
@davidkarlsen can you take a look when you have some free cycle? thanks!

davidkarlsen (approved):
lgtm

cep21:
This is a great PR for our CI/CD system.  Looking forward to getting it in a released version!

PR #429, opened 2022-04-20
refactor setup tools

cpanato:
- pin dependencies to be used for ci/release and remove form the go modules

as same did in https://github.com/helm/chart-releaser/pull/181


davidkarlsen (approved):
Lgtm

Issue #433, opened 2022-05-06
Upgrade tests is installing and upgrading the same chart version


paologallinaharbur:
**Is this a request for help?**:
No

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
Yes, unless I am misunderstanding the docs

**Version of Helm and Kubernetes**:
```
helm version
version.BuildInfo{Version:"v3.8.0", GitCommit:"d14138609b01886f544b2025f5000351c9eb092e", GitTreeState:"clean", GoVersion:"go1.17.6"}

 k version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:51:05Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.3", GitCommit:"816c97ab8cff8a1c72eccca1026f7820e93e0d25", GitTreeState:"clean", BuildDate:"2022-01-25T21:19:12Z", GoVersion:"go1.17.6", Compiler:"gc", Platform:"linux/amd64"}
```

**What happened**:
We maintain some charts, we modified by mistake a selector between two minor versions and the CI running ` ct install --namespace ct --upgrade ` did not catched the issue

**How to reproduce it** (as minimally and precisely as possible):
Clone a repository with a chart, modify a deployment label selector updating the version, run the tool

No error will be catched, but if you install the old version and the upgrade you would get the classical:
```
Error: UPGRADE FAILED: cannot patch "nri-prometheus" with kind Deployment: Deployment.apps "nri-prometheus" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/name":"nri-prometheus"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
``` 

**Anything else we need to know**:
I checked the code and I cannot understand the logic of the `--upgrade` function, but I checked and it is the very same of the first implementation, therefore I wonder how it is possible that it is a bug:

The tool states that is going to perform an upgrade test, but that is not the case! It upgrades to the same version `ct_previous_revision2130172757`
```
[...]

Issue #198, opened 2020-01-15
install --upgrade delete namespace at end of previous version install


n0rad:
Hi there,

I'm trying to use `install --upgrade` to validate that a chart is upgradable from one version to another.
At the end of the *previous* version installation, the namespace is deleted and a new one is created to install *current* version

I'm testing with CT latest container (2.4.0) against a private repository with changes on a branch.
from the logs changes are well detected : 

```
Testing upgrades of chart 'maxscale => (version: "0.1.1", path: "maxscale")' relative to previous revision 'maxscale => (version: "0.1.0", path: "ct_previous_revision535151147/maxscale")'...
```

But at the end of the *previous* install : 
```
...
Deleting release 'maxscale-2semyflfvv'...
>>> helm delete --purge maxscale-2semyflfvv
release "maxscale-2semyflfvv" deleted
Deleting namespace 'maxscale-2semyflfvv'...
>>> kubectl delete namespace maxscale-2semyflfvv --timeout 180s
namespace "maxscale-2semyflfvv" deleted
>>> kubectl get namespace maxscale-2semyflfvv
Namespace 'maxscale-2semyflfvv' terminated.
Testing upgrades of chart 'maxscale => (version: "0.1.1", path: "maxscale")' relative to previous revision 'maxscale => (version: "0.1.1", path: "maxscale")'...
>>> helm install maxscale --name maxscale-mgmh215ffh --namespace maxscale-mgmh215ffh --wait
...
```

PR #434, opened 2022-05-06
FIX: it should upgrade to new version instead of the old one

paologallinaharbur:
When testing the upgrade we should actually install the new version

Old behaviour (installing the same version, not catching the issue with selectors)
```
Testing upgrades of chart 'nri-prometheus => (version: "2.1.2", path: "charts/nri-prometheus")' relative to previous revision 'nri-prometheus => (version: "2.1.1", path: "ct_previous_revision2130172757/charts/nri-prometheus")'...
[...]
>>> helm install nri-prometheus-y2dumyevqw ct_previous_revision2130172757/charts/nri-prometheus --namespace ct --wait --values ct_previous_revision2130172757/charts/nri-prometheus/ci/test-lowdatamode-values.yaml
>>> helm upgrade nri-prometheus-y2dumyevqw ct_previous_revision2130172757/charts/nri-prometheus --namespace ct --reuse-values --wait
```

New Behaviour (installing the new version, catching the issue)
```
Testing upgrades of chart 'nri-prometheus => (version: "2.1.2", path: "charts/nri-prometheus")' relative to previous revision 'nri-prometheus => (version: "2.1.1", path: "ct_previous_revision78831704/charts/nri-prometheus")'...
[...]
>>> helm install nri-prometheus-avhvhn5q0a ct_previous_revision78831704/charts/nri-prometheus --namespace ct --wait --values ct_previous_revision78831704/charts/nri-prometheus/ci/test-lowdatamode-values.yaml
>>> helm upgrade nri-prometheus-avhvhn5q0a charts/nri-prometheus --namespace ct --reuse-values --wait
Error: UPGRADE FAILED: cannot patch "nri-prometheus" with kind Deployment: Deployment.apps "nri-prometheus" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/name":"nri-prometheus"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable

```

**Which issue this PR fixes** 

Fix https://github.com/helm/chart-testing/issues/433

Please let me know if I am misunderstanding the feature or the code (it is a vely likely possibility 😄 )

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -628,7 +628,7 @@ func (t *Testing) doUpgrade(oldChart, newChart *Chart, oldChartMustPass bool) er
 				return nil
 			}
 
-			if err := t.helm.Upgrade(oldChart.Path(), namespace, release); err != nil {
+			if err := t.helm.Upgrade(newChart.Path(), namespace, release); err != nil {
 				return err
 			}
 


paologallinaharbur:
If interested, it would make sense as well to add some tests to cover the bug (if that's a bug 😅  ). For example with two charts having as only difference the selectors, like:
```go
		{
			"change immutable deployment.spec.selector field in new version of the chart,
			"test_charts/simple-deployment",
			"test_charts/simple-deployment-different-selector",
			processError,
		},
```

---

Moreover, it would make sense to garbage collect the upgrade test `change immutable deployment.spec.selector field` since it covers https://github.com/helm/charts/issues/7726 that it is due to `apps/v1beta1`, but it using to do so a v1 chart, .

paologallinaharbur:
Hello @cpanato, 😄 
do you know who can we ask to in order to know if it is a bug of the `--upgrade` flag or not? 
I had the feeling that at least in our use case it is a possible source of false positives

LeoColomb:
I also believe it's a bug, and this fix seems relevant.
Should also close https://github.com/helm/chart-testing/issues/198

davidkarlsen (approved):

PR #445, opened 2022-06-08
feat(helm): Add support for OCI chart repositories

jcsf:
Since Helm 3.8 the support of OCI registries is no longer experimental and is enabled by default, this means that more and more people will be using registries to store charts. In order to support this type of repository, a small change is needed because OCI registries can't be added like "regular repositories", and need to use a different command "helm registry login".

Therefore, and since it was [already agreed](https://helm.sh/blog/storing-charts-in-oci/) that OCI repositories should start the URL with "oci://" this PR uses the prefix of the repository URL to select the right command to be able to add these type of repositories.

**Special notes for your reviewer**: I didn't added any automation test because didn't find any similar to what already was done, I tested this locally using a registry of mine.

Signed-off-by: João Fernandes <jcsf_1995@hotmail.com>

### DIFF
--- a/pkg/tool/helm.go
+++ b/pkg/tool/helm.go
@@ -16,6 +16,7 @@ package tool
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/helm/chart-testing/v3/pkg/exec"
 )
@@ -35,6 +36,13 @@ func NewHelm(exec exec.ProcessExecutor, extraArgs []string, extraSetArgs []strin
 }
 
 func (h Helm) AddRepo(name string, url string, extraArgs []string) error {
+	const ociPrefix string = "oci://"
+
+	if strings.HasPrefix(url, ociPrefix) {
+		registryDomain := url[len(ociPrefix):]
+		return h.exec.RunProcess("helm", "registry", "login", registryDomain, extraArgs)
+	}
+
 	return h.exec.RunProcess("helm", "repo", "add", name, url, extraArgs)
 }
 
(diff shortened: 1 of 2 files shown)


tuananh-finx:
can someone review this? we're adopting CI and currently running this custom build as workaround

tuananh:
bumping this up for visibility :)

jcsf:
Can a maintainer look at this? It is a really small review that helps a lot of users.

hungran:
+1 on this feature
https://github.com/pipe-cd/pipecd/issues/4588

cpanato:
@jcsf sorry for the delay here, can you rebase and fix the lint issues?

jcsf:
> @jcsf sorry for the delay here, can you rebase and fix the lint issues?

Sure, will work on this today and update the PR.

pkg/tool/helm.go
```diff
@@ -18,6 +18,8 @@ import (
 	"fmt"
 
 	"github.com/helm/chart-testing/v3/pkg/exec"
+
+	"strings"
```
cpanato, line 22: can you move this to close to the "fmt" part

should be like

```
import (
	"fmt"
	"strings"

	"github.com/helm/chart-testing/v3/pkg/exec"
)
```

cpanato (changes requested):

pkg/tool/helm.go
jcsf: Done

cpanato (approved):
thank you!

Issue #359, opened 2021-10-06
Improve error logs when a service is unavailable


goelankitt:
**Is this a request for help?**:
No
---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
Feature Request

I set the default value of a service type to "LoadBalancer". And hence that service would not become ready on the kind cluster. The ct tool eventually waits for the service to become ready and then times out. The error log does not show anything related to the actual failure i.e. the service not becoming ready. I was just wondering if it's possibly to improve the error log in this case by printing the service(or any other k8s resource) that is not ready.

You can check the CI logs here:
[Logs](https://github.com/deliveryhero/helm-charts/runs/3791289284)

**Version of Helm and Kubernetes**: 

**What happened**: The service was stuck in Pending state because it's type was "LoadBalancer" and ct failed with following error:
```
------------------------------------------------------------------------------------------------------------------------
 ✖︎ cortex-gateway => (version: "0.1.0", path: "stable/cortex-gateway") > Error waiting for process: exit status 1
------------------------------------------------------------------------------------------------------------------------
Error installing charts: Error processing charts
Error: Error installing charts: Error processing charts
Error: Process completed with exit code 1.
```

**What you expected to happen**: I expected it to print which service/pods were unavailable/not ready.

**How to reproduce it** (as minimally and precisely as possible): In any helm chart default values file set the type of service to LoadBalancer.

**Anything else we need to know**: No

PR #470, opened 2022-09-14
Fix parsing kubectl output with warnings

usmonster:
Exclude stderr (e.g. deprecation warnings) when parsing command output.

**What this PR does / why we need it**: It ensures that warnings or other output that goes to stderr won't be parsed when trying to list deployments, pods, etc. Today, this is not the case, and many  `ct` subcommands will fail in such situations (e.g. when the [gke-gcloud-auth-plugin](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke) isn't yet installed).

It's possible that the original implementation wanted to include stderr in the output for these commands for a reason, but this feels like a simple oversight.

It's also possible that there are other similar places in the code that require the same modification, so please feel free to update the PR or request these changes, which I will then apply. Thanks!

### DIFF
--- a/pkg/tool/kubectl.go
+++ b/pkg/tool/kubectl.go
@@ -146,7 +146,7 @@ func (k Kubectl) forceNamespaceDeletion(namespace string) error {
 }
 
 func (k Kubectl) WaitForDeployments(namespace string, selector string) error {
-	output, err := k.exec.RunProcessAndCaptureOutput("kubectl",
+	output, err := k.exec.RunProcessAndCaptureStdout("kubectl",
 		fmt.Sprintf("--request-timeout=%s", k.timeout),
 		"get", "deployments", "--namespace", namespace, "--selector", selector,
 		"--output", "jsonpath={.items[*].metadata.name}")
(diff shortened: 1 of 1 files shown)


usmonster:
Bump! @cpanato, @smlx—as recent committers on this file, can you please take a look as soon as you get a chance? 🙏 Thanks!

usmonster:
Hello, just checking in—@davidkarlsen, @scottrigby, do you think you can take a quick look at this fix and hopefully get it into the next version? Thanks again! 🙇

davidkarlsen:
I'm a bit uneasy to merge it straight as I suspect some unexpected side-effects, but I'd like @unguiculus comments as the original author. Do you remember the reasoning between getting both streams here?

unguiculus:
Maybe we should adjust `RunProcessAndCaptureOutput` so that it also logs the proc's stderr to stderr, so we don't swallow any errors. Otherwise, I'd be fine with this change.

usmonster:
Thanks @unguiculus! I like your suggestion as a separate PR, which could also help with #359.

Regarding @davidkarlsen's concerns, there should be no unwanted side effects since we're now only filtering stderr output that isn't currently used or expected in these cases.

Please feel free to merge if there's nothing else to change, and thanks again for the review!

usmonster:
Just a friendly bump to @davidkarlsen & @scottrigby. 🙂

Do you need anything else from me before this can be merged? Thanks again!

davidkarlsen (approved):
lgtm

cpanato (approved):

PR #474, opened 2022-09-19
fix: using helm version --short resulting in an invalid semantic versioning

### LINKED ISSUE #467: Helm version verification fails when using RedHat's Helm build
Hello there,

My infrastructure team recently updated the docker image we use for most of our CI/CD and substituted Helm's official binary for RedHat's. (We use Redhat's Openshift)
Since then, CT has been unusable and returns `Invalid Semantic Version` errors.

We could have it just rolled back, but I thought I'd have a look into it...

## How to reproduce

- Grab any of the RedHat's build from their [repository](https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/helm/)

For example at the time of this writing the latest version returns:
 `version.BuildInfo{Version:"v3.9.0+3.el8", GitCommit:"e09b16a5119e20607a4a7ae9884a96331ffff1de", GitTreeState:"clean", GoVersion:"go1.17.7"}`

- Make this version the one resolved by your path

- Run, for example, `ct lint`

This will produce the following output:

```
> ct lint
Linting charts...
Error: Invalid Semantic Version
Invalid Semantic Version
```

### What is happening?

The error occurs here:
https://github.com/helm/chart-testing/blob/1d3feac8e5ca55ccf11521ff297ccee7f09aa749/pkg/chart/chart.go#L279-L282

While debugging, it appears that the `versionString` passed is equal to `v3.9.0+3.el8+ge09b16a`.

Which indeed does not match the SemVer regex found [here](https://github.com/Masterminds/semver/blob/b8121eb457fd2944a4bf285e3197f729f5595b95/version.go#L42-L44): 
`^v?([0-9]+)(\.[0-9]+)?(\.[0-9]+)?(-([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))?$`

Looking at https://semver.org/, multiple `+` does not seem to be a valid syntax.

### Why is the version rendered as such?

Looking at the following:
https://github.com/helm/chart-testing/blob/1d3feac8e5ca55ccf11521ff297ccee7f09aa749/pkg/tool/helm.go#L92-L94

Helm's short version is used, which is rendered by Helm as seen [here](https://github.com/helm/helm/blob/8199db309ae73d745b983fb7ea2bfdd944a834ae/cmd/helm/version.go#L94-L103), by concatenating the version followed by `+g` and then the first 7 characters of the GitCommit.

As stated above, the full version is  `version.BuildInfo{Version:"v3.9.0+3.el8", GitCommit:"e09b16a5119e20607a4a7ae9884a96331ffff1de", GitTreeState:"clean", GoVersion:"go1.17.7"}`, hence the incorrect version format.

## How to fix this?

This is not for me to say, obviously. :)

The options I identify, are the following:

### Change chart-testing's Helm.Version() implementation

One thing that could be done is to parse Helm's long version to only retrieve the `Version` part instead of using Helm's short version.
It seems like only the Major component is used, to make sure that Helm's version is > 3

\+ Easy and quick to fix, only involves changes in CT
\- Does not fix the fact that Helm returns an improper SemVer with the `--short` flag

### Change Helm's formatVersion to handle the special case of the Version already containing a "build" component

A fix could be operated in Helm's [formatVersion function](https://github.com/helm/helm/blob/8199db309ae73d745b983fb7ea2bfdd944a834ae/cmd/helm/version.go#L94-L103) to handle this case of the Version already containing a "build" component (`(\+([0-9A-Za-z\-]+(\.[0-9A-Za-z\-]+)*))`).  
It could for example, in this case, use `.` as a separator instead of a `+`, as it is supported by the SemVer spec.

\+ Properly fixes the problem
\- Requires to make changes in Helm and then have to wait for new version and RH build

My preference goes for the second one, but I'd love to have your take on this!

I'd happily open a PR if needed! :)

tisutisu:
Signed-off-by: Sushanta Das <sushanta.das.ju@gmail.com>

On using `helm version --short` resulting in invalid sematic version if the version includes a build part in the version string.
Instead we can use the `helm version --template {{ .Version }}` to get the helm version. Please see the issue for more information about the problem.

fixes https://github.com/helm/chart-testing/issues/467

### DIFF
--- a/pkg/tool/helm.go
+++ b/pkg/tool/helm.go
@@ -90,5 +90,5 @@ func (h Helm) DeleteRelease(namespace string, release string) {
 }
 
 func (h Helm) Version() (string, error) {
-	return h.exec.RunProcessAndCaptureStdout("helm", "version", "--short")
+	return h.exec.RunProcessAndCaptureStdout("helm", "version", "--template", "{{ .Version }}")
 }


sabre1041 (approved):
LGTM

cpanato (approved):
thank you!

PR #510, opened 2023-01-06
(Breaking change) chore: change default target-branch to main

fty4:
**What this PR does / why we need it**:

Because the default branch on GitHub is [`main`](https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/) and not longer `master` this should also be the default to chart-testing.

**Which issue this PR fixes**:

fixes #376 

**Special notes for your reviewer**:

BREAKING CHANGE:
If e.g an action or workflow is not using an explicit specified `target-branch` the job might will fail after this change.

But after this change new users does not have the problem that a wrong branch is used.
For example I was suffering with the error message `exit status 128`- see #330 because the default branch was not `main`.

PR #511, opened 2023-01-06
feat: verify if targetBranch is present in git repo

### LINKED ISSUE #330: Add better error message if "target branch" is missing - exit status 128
**Is this a request for help?**: no

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

**Version of Helm and Kubernetes**: `helm 3.4.0`

**What happened**:

When running `ct list-changed` with the default target-branch (`master`) - if `master` does not exist (i.e. GitHub now defaults to `main`), then you get the following in GitHub Actions (via the [chart-testing-action](https://github.com/helm/chart-testing-action)):

```
Error: Error running process: exit status 128
```

This is very tricky to debug without a dev environment that mirrors the GitHub Action (unfortunately I had a `master` branch locally so things would not fail locally). Mirroring the GitHub Action environment is not trivial - a bit of logging or a cleaner error would go a long ways to making this easier to debug!

I eventually figured this out by reasoning from this issue: https://github.com/helm/chart-testing/issues/186

```
ERROR: `--target-branch=master` does not exist
```
(Or infer which branch should be referenced from the git repository)

```
# in a git repo with no master branch
ct list-changed
```

**Anything else we need to know**:

fty4:

If a targetBranch is not present in the git repository the `ct list-changed` command will fail. The error message for this is not really meaningful.

This PR will validate if the branch exists and throw a error message which the user will understand.

Because the default branch in GitHub changed from `master` to [`main`](https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/) this is a common problem users are facing.

I've built the chart-testing and tested it with master branch as follows:

```bash
$ ct list-changed
Error: targetBranch 'master' does not exist
targetBranch 'master' does not exist
```

**Which issue this PR fixes**:

closes #330

- I am not sure if I should add a test here..
  other functions are not really care about it as `FileExistsOnBranch` for example:
  https://github.com/helm/chart-testing/blob/89d34943ed04b8bacad9693c6130f28a6fb3a6c7/pkg/chart/chart_test.go#L30-L32

- related to #510

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -708,7 +708,7 @@ func (t *Testing) computeMergeBase() (string, error) {
 	}
 
 	if !t.git.BranchExists(t.config.TargetBranch) {
-		return "", errors.New(fmt.Sprintf("targetBranch '%s' does not exist", t.config.TargetBranch))
+		return "", fmt.Errorf("targetBranch '%s' does not exist", t.config.TargetBranch)
 	}
 
 	return t.git.MergeBase(fmt.Sprintf("%s/%s", t.config.Remote, t.config.TargetBranch), t.config.Since)
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -49,6 +49,8 @@ const maxNameLength = 63
 //
 // ValidateRepository checks that the current working directory is a valid git repository,
 // and returns nil if valid.
+//
+// BranchExists checks whether a given branch exists in the git repository.
 type Git interface {
 	FileExistsOnBranch(file string, remote string, branch string) bool
 	Show(file string, remote string, branch string) (string, error)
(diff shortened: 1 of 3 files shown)


davidkarlsen:
> * I am not sure if I should add a test here..
>   other functions are not really care about it as `FileExistsOnBranch` for example:
>   https://github.com/helm/chart-testing/blob/89d34943ed04b8bacad9693c6130f28a6fb3a6c7/pkg/chart/chart_test.go#L30-L32
> * related to [chore: change default target-branch to main #510](https://github.com/helm/chart-testing/pull/510)

if you look in `chart_test.go` there is a mock, if fails on the missing method now. Maybe it is possible to add a test there. At least the method needs to be stubbed out.

fty4:
@davidkarlsen thank you for your comment.

I've now mocked the test for func `chart.BranchExists()`.
I guess at this moment it would not be that easy to add a real test.
Also I was wondering where to put this test - there is no order in the file, right?

The linter also suggested to change the error msg generation.

_Note: required force-push because of missing DCO_

PR #620, opened 2024-01-08
feat(docker): Bump dependencies in Dockerfile

meyfa:

This patch bumps the versions of pip, kubectl, helm, and yamllint.

I have run build.sh locally to verify it still builds, and all the tools are present and working at the proper versions when running the image with Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,10 +9,10 @@ RUN apk --no-cache add \
     py3-pip \
     py3-wheel \
     python3 && \
-    pip install --upgrade pip==21.0.1
+    pip install --upgrade pip==23.3.2
 
 # Install a YAML Linter
-ARG yamllint_version=1.29.0
+ARG yamllint_version=1.33.0
 LABEL yamllint-version=$yamllint_version
 RUN pip install "yamllint==$yamllint_version"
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -9,10 +9,10 @@ RUN apk --no-cache add \
     py3-pip \
     py3-wheel \
     python3 && \
-    pip install --upgrade pip==21.0.1
+    pip install --upgrade pip==23.3.2
 
 # Install a YAML Linter
-ARG yamllint_version=1.29.0
+ARG yamllint_version=1.33.0
```
cpanato, line 15: there is any breaking changes that migth affect the users?

cpanato (changes requested):
thanks
please sign the dco

Dockerfile
meyfa: As far as I can tell, no. Relevant updates:

- one new rule (should be irrelevant unless users configure this, so not breaking)
- new rule options (also not breaking)
- fixed false-positives (meaning less errors, therefore not breaking)

Ref: https://github.com/adrienverge/yamllint/blob/master/CHANGELOG.rst

cpanato: thanks for checking

cpanato (approved):

meyfa:
@cpanato Thanks for the review & merging! Is there any timeline for a new release?

cpanato:
@meyfa trying, need to find a bit more time

PR #531, opened 2023-03-11
allow reuse-values on upgrade to be configurable

joejulian:
**What this PR does / why we need it**:

many charts add new values between chart versions. When using the reuse-values flag, those charts fail because reuse-values ignores the default values file.

By making this configurable, it allows the chart maintainer to determine whether it's appropriate for their chart to add new values when upgrading.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #525 

**Special notes for your reviewer**:

PR #652, opened 2024-05-26
feat: options to configure chart upgrade strategy

### LINKED ISSUE #525: helm upgrade uses `reuse-values` by default
**Is this a request for help?**:

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):

BUG REPORT

**Version of Helm and Kubernetes**:

version.BuildInfo{Version:"v3.11.0", GitCommit:"472c5736ab01133de504a826bd9ee12cbe4e7904", GitTreeState:"clean", GoVersion:"go1.19.5"}

ct version 2.7.1

**What happened**:
Upgrade failed when adding a new subchart in the new version because the new default chart values are ignored when reuse-values is specified. This caused the overrides in the new parent chart's values to be omitted and the subcharts defaults were used instead. As those overrides are required for the subchart to be a subchart, this caused the upgrade to fail.

When testing without `reuse-values` the upgrade is performed correctly.

Since using that flag is not a commonly documented way of upgrading charts, it should not be the default.

I expected the common `helm upgrade <release> <chart> --wait` to be executed.

**Anything else we need to know**:

tthvo:

Added options (i.e.  `--upgrade-strategy <option>`) to configure upgrade strategy for chart values. This allows maintainers to choose whichever upgrade paths align with their charts.

Without specifying the flag, default to `reuse-values` for backwards compatibility.

Follow up for #531 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #525 

Sorry if it duplicates any in-progress work. Please let me know what you think :D

### DIFF
--- a/ct/cmd/install.go
+++ b/ct/cmd/install.go
@@ -79,6 +79,9 @@ func addInstallFlags(flags *flag.FlagSet) {
 		(e.g. "--set=name=value"`))
 	flags.Bool("skip-clean-up", false, heredoc.Doc(`
 		Skip resources clean-up. Used if need to continue other flows or keep it around.`))
+	flags.String("upgrade-strategy", "reuse-values", heredoc.Doc(`
+		When --upgrade is used, this flag configures the upgrade strategy for chart values.
+		Available options: reset-values, reuse-values, reset-then-reuse-values`))
 }
 
 func install(cmd *cobra.Command, _ []string) error {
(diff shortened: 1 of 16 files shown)


tthvo:
Hi @cpanato, would you have time to look at this? Sorry, I can't add/tell who to request reviews :D

TheChef23:
This is great, this would fix the issues I'm having where the chart upgrade path cannot be tested because of the adding of new default values which don't get included in the upgrade.

tthvo:
Hi, would anyone mind having a look at this PR? I think it would be useful for testing with different upgrade paths.

tthvo:
I haven't looked into this for long. Let me rebase the PR and reopen it this week if it is still a good solution 😅

PR #713, opened 2025-02-08
Fix `helm-extra-set-args` in YAML configuration

arbourd:
**What this PR does / why we need it:**

Allows the usage of `helm-extra-set-args` key from a `ct.yaml` and not just command-line flags. 

Currently, `helm-extra-set-args` is only available as a command-line flag, despite being part of the config object. This does not appear to be documented.

This change sources the value `extraSetArgs` via the `config` object rather than the cmd.Flags() only.

This PR adds the configuration option but it is never used: https://github.com/helm/chart-testing/pull/697

### DIFF
--- a/ct/cmd/install.go
+++ b/ct/cmd/install.go
@@ -96,11 +96,7 @@ func install(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("failed loading configuration: %w", err)
 	}
 
-	extraSetArgs, err := cmd.Flags().GetString("helm-extra-set-args")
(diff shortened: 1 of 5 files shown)


ryan-surname-p (approved):

arbourd:
@unguiculus Is it possible to get this fix in, please?

cpanato:
@arbourd can you rebase?

arbourd:
@cpanato Done.

arbourd:
No, thank you! Appreciate it.

PR #755, opened 2025-07-24
feat: skip values.yaml yamllint when file is absent

### LINKED ISSUE #748: Feature request: skip `values.yaml` yamllint when it is absent
Currently, `ct lint` fails whenever `values.yaml` is missing in a chart.

Sometimes there is simply no need to provide `values.yaml` (e.g. when installing manifests that don't need to be configurable).

**Expected**: `ct lint` should not fail when `values.yaml` is absent.

**Observed**:

```
% ct lint
[...]
[Errno 2] No such file or directory: 'helm/mychart/values.yaml'

------------------------------------------------------------------------------------------------------------------------
 ✖︎ mychart => (version: "1.2.2", path: "helm/mychart") > failed waiting for process: exit status 255
------------------------------------------------------------------------------------------------------------------------
Error: failed linting charts: failed processing charts
failed linting charts: failed processing charts
```

When running with `debug: true`, I see:

```
>>> yamllint --config-file .yamllint.yaml helm/mychart/Chart.yaml
>>> yamllint --config-file .yamllint.yaml helm/mychart/values.yaml
[Errno 2] No such file or directory: 'helm/mychart/values.yaml'
```

So the problem is that it attempts to run `yamllint` in the non-existing `values.yaml`.

I have `validate-yaml: true` in my `ct` config.

The current workaround is to create an empty `values.yaml` file in the chart, just for the sake of pleasing `ct lint`.

thiagowfx:
Resolve issue where ct lint fails when values.yaml is missing from a
chart directory. Validate values.yaml with yamllint if the file actually
exists, allowing charts without configurable values to pass linting
successfully.

Testing:

- All existing tests continue to pass
- New test specifically validates that yamllint is only called once (for
  Chart.yaml) when values.yaml is missing
- Code passes go vet, go fmt, and golangci-lint checks

Behavior:

- Before: ct lint would fail if values.yaml was missing from a chart,
  even when validate-yaml: true was set
- After: ct lint gracefully skips yamllint validation for values.yaml
  when the file doesn't exist, allowing charts without configurable
  values to lint successfully

Fixes #748


pkg/chart/chart_test.go
```diff
@@ -396,6 +396,53 @@ func TestLintYamlValidation(t *testing.T) {
...
+
+		var suffix string
+		if validate {
+			suffix = "with-validation"
+		} else {
+			suffix = "without-validation"
+		}
+
+		testCases := []testData{
+			{fmt.Sprintf("lint-no-values-yaml-%s", suffix), "testdata/no_values_yaml", true},
```
cpanato, line 425: should add with values as well

thiagowfx:
Thanks for the feedback. Aiming to update it this month.

PR #806, opened 2026-01-29
feat: add `--helm-uninstall-extra-args` flag

### LINKED ISSUE #737: --helm-extra-args is applied to all helm commands
I'd like to use `ct install` in helm dry-run mode. Trying to do that using `ct install --debug --helm-extra-args="--dry-run=server --debug" --target-branch main` will fail because the arguments are also applied to other helm commands like `helm test` and `helm uninstall` which don't recognise the dry-run argument. 
Maybe an option `--helm-install-extra-args` is needed?

pmalek:

**Which issue this PR fixes** 

Fixes #737 


pmalek:
Still waiting for review on this one.

